PR #3, opened 2025-10-29
rewrite still with libxev and a single daemon

neurosnap:


x1f9-ricon:
Addressed both review items in ad62186:

1. **TTY side-channel**: `writeTable` now takes `use_color: bool` as a parameter. Caller in main.zig passes `isTty(stdout)`. Function is now testable and correct for non-stdout writers.

2. **JSON string escaping**: Added `writeJsonString` helper that handles \", \\, \n, \r, \t, and RFC 8259 control characters (`\u00XX` for bytes < 0x20). All string fields now go through it — name, status, age, start_dir, cmd.

Added tests for the escaping helper and for JSON output with quotes in session names. All 48 BATS + unit tests pass.

PR #5, opened 2025-12-02
add nix flake

brittonr:
add basic nix flake using zig2nix.

### DIFF
--- a/build.zig.zon
+++ b/build.zig.zon
@@ -25,7 +25,7 @@
     .fingerprint = 0x28aad87005052b4e, // Changing this has security and trust implications.
     // Tracks the earliest Zig version that the package considers to be a
     // supported use case.
-    .minimum_zig_version = "0.15.1",
+    .minimum_zig_version = "0.15.2",
     // This field is optional.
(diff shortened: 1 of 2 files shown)


neurosnap:
Hello!  I don't use nix so it's unclear to me what needs to happen when I change my zig dependencies.  How do we keep the lock file updated?

brittonr:
Ha yes that would be good to know. To update the `build.zig.zon2json-lock` file that nix makes, you'd just run `nix run github:Cloudef/zig2nix -- zon2lock build.zig.zon` (assuming nix is installed on the system) when dependencies are updated. I'm not super familiar with zig (this project just solves a problem I've had for a while), but I understand this lock file isn't very standard in zig, but more so a nix quirk. If you'd like, I can look into setting up an action to update this automatically.

AkechiShiro:
I can also help, regarding the automatic action and maintenance @brittonr @neurosnap 

As I also would like this PR to be merged and a `default.nix` (flake-compat aka flakeless usage) be added if possible

neurosnap:
@AkechiShiro hey thanks! I'm going to merge this PR now and I'll create an issue about nix package maintenance.

PR #21, opened 2025-12-07
Add Homebrew formula and release script update

GabrielDougherty:
This PR adds a Homebrew formula so that zmx can be installed with `brew install`. It also updates the release step of the build.zig to update the formula with links to the new package version, which means you would have to commit the result after performing a release. If that' not your desired workflow, I'm open to doing something different.


neurosnap (commented):
Thanks for your contribution!

Is this the flow for a release?:

```bash
zig build release
zig build upload
git add Formula/zmx.rb && git commit
git push
```

I'm also not a huge fan of requiring `python3` in the build, can you think of any alternatives?  It's not a blocker since I do really want this merged, just looking for ideas as potential follow-ups.

neurosnap:
It looks like the official docs recommend a `{user}/homebrew-tap` repo instead of having it inside the upstream repo: https://docs.brew.sh/How-to-Create-and-Maintain-a-Tap

I'll probably follow this guide to create the tap but will be using your contribution as a template.  Thanks!

neurosnap:
https://github.com/neurosnap/homebrew-tap

PR #27, opened 2025-12-19
Use XDG_RUNTIME_DIR for the socket directory if defined

tristan957:
XDG_RUNTIME_DIR[0] is a file path for user-specific runtime files, like sockets.

It will typically be defined on Linux. I can't say for sure on BSDs, and I know for certain it is not defined on macOS.

Link: https://specifications.freedesktop.org/basedir/latest/ [0]

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 Use spec: https://common-changelog.org/
 
+## Unreleased
+
+### Added
+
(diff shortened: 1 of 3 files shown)


tristan957:
I can make sure it is used only on linux if requested.

neurosnap:
Thanks! af42d4ec49c76e4484c5c1071f8203b0f8204a68

PR #47, opened 2026-01-20
fix: SIGHUP instead of SIGTERM for kill

cwelsys:
Interactive shells (often) ignore SIGTERM.

### DIFF
--- a/src/main.zig
+++ b/src/main.zig
@@ -241,8 +241,8 @@ const Daemon = struct {
 
     pub fn handleKill(self: *Daemon) void {
         std.log.info("kill received session={s}", .{self.session_name});
-        posix.kill(self.pid, posix.SIG.TERM) catch |err| {
-            std.log.warn("failed to send SIGTERM to pty child err={s}", .{@errorName(err)});
+        posix.kill(self.pid, posix.SIG.HUP) catch |err| {
+            std.log.warn("failed to send SIGHUP to pty child err={s}", .{@errorName(err)});
         };
         self.shutdown();
     }


cwelsys:
Noticed a bunch of orphaned processes long after zmx killing them.

This PR serves as both a question and a fix. Was there a reason I am not accounting for that TERM is sent instead of HUP?

Anyways thank you for the project, it rules. cheers.

neurosnap:
Thanks for the PR.  I'm pretty sure `SIGTERM` makes the most sense since it's supposed to denote a graceful shutdown notice. I think the issue is likely that we are not forwarding the SIGTERM to children processes.

I pushed a fix to `main` if you could test it out and let me know if you notice any other orphaned processes that would be great!

https://github.com/neurosnap/zmx/commit/377c062aec0361f7ae6588ce70b6d255f2d1a2f0

cwelsys:
Checking it out now, 

Still getting orphans. Now i'm reading the [bash manual re: signals](https://www.gnu.org/software/bash/manual/html_node/Signals.html).

I'm using zsh, but this behavior seems consistent across the shells. (ive only actually ran tests with bash/zsh)

test was ~:
zmx a test
echo $$ && echo $ZMX_SESSION
zmx kill test
zmx list
session_name=test	status=Timeout	(cleaning up)
pgrep zsh
see pid identified above.

Bash manual states "in the absence of any traps, it ignores SIGTERM" Could set up a trap when spawning the child, then they might listen. 

Sorry for the blog, hope this was useful in some way.

neurosnap:
Hmm darn. I'll read that post and think about this some more. So I can test locally, are you running zmx kill inside the session that you're killing?

cwelsys:
> Hmm darn. I'll read that post and think about this some more. So I can test locally, are you running zmx kill inside the session that you're killing?

I appreciate it. and no, ran it from another window, but i have tried running it from inside the session with the same result.

neurosnap:
Alright, thanks for link that was helpful.  My current plan is to run `SIGHUP` + timeout + `SIGKILL` since users could run a program that doesn't handle SIGHUP properly.  I'll push some code later today after I have confirmed the behavior I want.

Thanks for your patience, I learned something new today.

cwelsys:
No worries re: patience, thank you for looking into it. i'll 'close with comments' this PR since its not really a proper fix.

Issue #39, opened 2026-01-06
Native SSH support


joshbainbridge:
More of a discussion topic than an issue. But have you considered native SSH support?

Personally, most of my sessions are remote, and I have been using the SSH config trick with different tools. However, it has some limitations. One is that you can't see the available sessions you already have live. The other is that you can't do any other operation other than attach. And it complicates the SSH config.

I had a quick play at making the zmx commands SSH aware. It is very simple. This follows the same design as other standard Unix tools like scp or rsync, and simplifies the completion logic.

So instead of:

``` bash
ssh -qt remote-server zmx attach fizz
```

You can call:

``` bash
zmx attach remote-server:fizz
zmx kill remote-server:fizz
# etc
```

I also tried out a `sync` command to enable zmx on the remote server. SSH config becomes very simple:

``` ssh
Host remote-server
    HostName 123.456.789
```

Here's a quick video showing it in action:

https://github.com/user-attachments/assets/a2936dda-ca07-4157-9df4-d59cec2b8a3c

neurosnap:
This is really cool and I do like the features you've presented here.  I'm concerned about complexity.  I do like the autocomplete and would almost certainly use it if I could but it doesn't feel that much different from `ssh {server} zmx`

joshbainbridge:
Apologies for the delay - it's been a busy few weeks. That's a fair point. There's a good chance this might just be too much added complexity for the payoff. Just experimenting. I really appreciate how minimal zmx is as a project.

The two main advantages over direct SSH are that you can tie it into the sync command and adapt the server's PATH to pick up your synced binary, plus you get remote autocomplete. But both are more nice-to-haves rather than essential features.

I may continue using this on my fork / clean it up a bit. If that then reaches a point where everything is as simple as can be, I'll share a link to the changes so you can get a measure of the complexity cost.

Related side note, it looks like there might be an opportunity to reduce complexity around the arg passing once [this](https://codeberg.org/ziglang/zig/issues/30677) proposal gets implemented and added to the std library. Just noticed that the other day.

neurosnap:
If you have a working solution let me know and I'll take a look at your fork.  The SSH workflow is a first-class citizen for this project so it's not like it is completely out-of-scope, but I do worry about code complexity.  I want to keep this project nice and small with as few configuration options as possible.

neurosnap:
After thinking about it some more I think this feature will add too much complexity to the project and feels a little out of scope.  Closing for now but happy to continue chatting about it

Issue #37, opened 2026-01-02
Adding auto-complete for commands and session names


joshbainbridge:
Would you consider integrating shell support for auto-completion? It could be limited to a few common shells such as bash, zsh and fish.

This can provide a couple benefits. It reduces the amount of typing needed when running a command. But perhaps more importantly, auto-completion also tends to list the available options (at least it does in zsh).

I often find that useful, especially when dealing with remote sessions over SSH, as I can't always recall what sessions might already exist.

If you think this would sit within the scope of the project, I did a quick POC on what that could like:

https://github.com/neurosnap/zmx/compare/main...joshbainbridge:zmx:shell-auto-completion

I've also done a short example video of this in action:

https://github.com/user-attachments/assets/8402dc35-7380-4424-9079-51fedbbf71d2

(the bell at the end shows that no sessions are available for auto-completion)

In the video, the SSH config would have looked something like:

```ssh
Host remote-server
    HostName 123.456.789
    RequestTTY yes
```

and the auto-completion needed this line in my `.zshrc`:

```zsh
eval “$(zmx completions zsh)"
```

If you'd like, I'd be happy to make a PR and move that forward. But no worries if you feel this isn't in scope or should be done differently :)

neurosnap:
I am interested in supporting auto-complete.  I wonder if there's a way to get this to work without making it a subcommand.  Other projects will include a `contrib/` folder that includes the completions: https://github.com/biox/pa/blob/main/contrib/pa-completion-new.bash

I don't want this to a be burden for users so if that means having a subcommand then that's fine.  I just want to explore other options before committing to it.

joshbainbridge:
That would be a reasonable option. It would keep the subcommand list that bit simpler.

Main drawback is maybe this would create a need for some form of installation or setup? You no longer have a single binary, but multiple files that need to be copied to a location, and then also cleaned up afterwords on uninstall. Binary and the files could also get out-of-sync.

Having a subcommand means a user can dynamically load the completions, without needing to keep files on disk. It does mean however they will be invoking zmx at the beginning of a new shell:

```bash
eval “$(zmx completions bash)"
```

Alternatively, a user would also have the option to write the completions to a file using the subcommand, like so:

```bash
zmx completions bash > ~/.local/share/bash-completion/completions/zmx
```

I think the dynamic option is the least friction for users. Pros and cons on both sides. Which direction are you leaning?

[...]

PR #48, opened 2026-01-21
Add shell auto-completion support

joshbainbridge:
Implements auto-completion for bash, zsh, and fish shells as discussed in #37.

**Changes:**
- Add `completions` command that outputs shell-specific completion scripts
- Add `--short` flag to `list` command for machine-readable output (used by completions)
- Update README with setup instructions

**Usage:**
```bash
# Add to .bashrc, .zshrc, or config.fish
eval "$(zmx completions bash)"
```

Auto-completion now works for:
- Subcommands (attach, run, list, etc.)
- Session names (from active sessions)
- Command flags and options

**Note:** I've left the SSH remote auto-completion (as seen on #37) out of this for now. It meant overriding the system SSH completion function, which could be seen as a bit heavy handed. However I can add that back if preferable. Or it would become a moot point if we opted to go with #39.

### DIFF
--- a/README.md
+++ b/README.md
@@ -137,6 +137,39 @@ POWERLEVEL9K_RIGHT_PROMPT_ELEMENTS+=my_zmx_session
    style = 'plain'
 ```
 
+## shell completion
+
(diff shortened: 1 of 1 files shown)


neurosnap:
Made a couple tweaks and merged, thanks!

PR #58, opened 2026-02-20
docs: add zmosh to community tools

laudney:
Hey neurosnap! First off, just want to say huge thanks for zmx. The architecture is brilliant—especially the decision to keep the daemon simple and use libghostty-vt. It made building zmosh a breeze.

I've been working on a fork called zmosh that takes zmx and adds encrypted UDP auto-reconnect (like mosh/ET) for flaky remote connections. It uses your exact daemon/session logic but wraps the client transport in XChaCha20-Poly1305.

Thought I'd add a link to it in the community tools section in case anyone else needs the roaming capability. Thanks again for the solid base!

### DIFF
--- a/README.md
+++ b/README.md
@@ -355,3 +355,4 @@ dtach is a program written in C that emulates the detach feature of screen, whic
 ## community tools
 
 - [zsm](https://github.com/mdsakalu/zmx-session-manager) — TUI session manager for zmx. List, preview, filter, and kill sessions from an interactive terminal UI.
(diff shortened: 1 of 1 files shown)


neurosnap:
Wow this is really cool!

ccoVeille:
How zmosh is different from zmx as a fork that the features you added cannot be added to zmx ?

Do you consider zmosh as done? And so, you won't add features but keep merging the zmx changes unto your fork?

I'm asking because while I like the idea of having the feature present in mosh, I would prefer the mosh features to be in zmx

Is there anything that prevent it? Putting aside the work to find solution to make them both work together

PR #60, opened 2026-02-21
fix(attach): use alt-screen to preserve outer shell on detach

laudney:
## Summary

- Enter alternate screen (`\e[?1049h`) before attaching instead of just clearing the screen
- Detach now automatically restores the prior outer shell content (cursor position + visible lines)
- Previously, attaching would destroy the outer terminal buffer, losing whatever was on screen before the session

## Test plan

- [ ] `zmx attach` to a session, run some commands, detach — outer shell should be fully restored
- [ ] Verify the session content renders correctly inside the alt-screen

### DIFF
--- a/src/main.zig
+++ b/src/main.zig
@@ -959,10 +959,10 @@ fn attach(daemon: *Daemon) !void {
 
     _ = c.tcsetattr(posix.STDIN_FILENO, c.TCSANOW, &raw_termios);
 
-    // Clear screen before attaching. This provides a clean slate before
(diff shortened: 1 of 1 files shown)


neurosnap:
We used to do this and then reverted because we noticed bugs related to using alt-screen for attach. Those could have been red herrings. I'll try to find the commit where we go from alt to primary for attach.

laudney:
Yeah, this is no good; alt-screen has no scroll buffer...

PR #61, opened 2026-02-21
fix(terminal): preserve scrollback through clear and fix cursor race on re-attach

### LINKED ISSUE #31: On reattach the active line is in the middle of the screen
Sometimes on reattach over ssh, the buffer is restored but the active line and cursor is placed in the middle of the screen. 

https://github.com/user-attachments/assets/edbf4aa5-79ca-4b72-a54c-21fdb8cd993b

laudney:
Fixes #31.

## Summary

**Fix 1 — Scrollback preservation through `clear`**

When the user runs `clear` (or any program that sends `ESC[2J`), the visible screen content was silently discarded — not moved to scrollback. On re-attach, the server serialized the full terminal state and sent it to the new client, which caused the entire history buffer to be visually rendered to the screen (scrolled through the viewport) before settling at the correct cursor position.

Two things now happen when `ESC[2J` is detected on the primary screen:
- `scrollClear()` is called on the server-side VT, moving the visible content into the scrollback pages before the erase
- `ESC[22J` is prepended to the client output stream so the client's terminal also pushes its current visible content to scrollback before clearing

**Fix 2 — Two-phase re-attach serialization**

The previous single-phase serialization (`content = null`, full terminal) emitted the entire history as a flat VT stream. The new client would render it by scrolling thousands of lines through the viewport — the entire history buffer visually printed to screen — before the active content settled at the right position.

Serialization is now split into two phases:
- **Phase 1:** history content only (no cursor, no modes). After emitting, `ESC[999;1H` + N newlines push the rendered content into the client's scrollback buffer without it remaining on screen
- **Phase 2:** active viewport only, with cursor position and full terminal modes

**Fix 3 — Cursor race on re-attach** (original scope of #31)

- Serialize **before** resize so the snapshot captures the pre-reflow cursor position
- `client.initialized` flag gates PTY broadcast and `POLL.OUT`, preventing live PTY output from reaching the client before its `Init` is processed
- Stale pre-Init buffered output is cleared before appending the snapshot
- Replies queued during `POLL.IN` (Init, Run, History, Info) are flushed immediately rather than waiting for the next poll cycle

## How to reproduce — scrollback dump bug (Fix 1 + 2)

```
zmx a repro
cat /usr/share/dict/words    # or any file that fills several screens
clear
```

Detach with `Ctrl+\`, then attach a second client in another terminal:

```
zmx a repro
```

**Old behaviour:** the entire file content scrolls rapidly through the new client's viewport before the cursor settles at the shell prompt. The history buffer is visually printed to screen on attach.

**Fixed behaviour:** the new client immediately shows the post-clear shell prompt; the file content is in the scrollback buffer and reachable by scrolling up.

## How to reproduce — cursor race bug (Fix 3)

1. Start with a wide terminal:
   ```
   stty rows 30 cols 140
   zmx a repro
   ```

2. Inside the session, generate long wrapped lines:
   ```
   for i in $(seq 1 160); do printf "%03d %s\n" "$i" "$(printf 'x%.0s' $(seq 1 220))"; done
   ```

3. Detach with `Ctrl+\`.

4. Shrink terminal:
   ```
   stty rows 30 cols 60
   ```

5. Reattach:
   ```
   zmx a repro
   ```

**Old behaviour:** session content is present but the cursor/active line appears mid-screen with text rendered below it. Noisy environments (SSH, active prompts) increase the likelihood.

**Fixed behaviour:** cursor stays at the bottom of the restored content after resize.

## Test plan

- [x] `cat large_file` → `clear` → re-attach: new client shows clean prompt, history accessible via scroll
- [x] Verify first attach (new session) works cleanly — no snapshot interference with shell init
- [x] Re-attach after resize restores content with cursor at the correct position
- [x] Utility clients (`zmx run`, `zmx history`) receive no spurious PTY output

### DIFF
--- a/src/main.zig
+++ b/src/main.zig
@@ -62,6 +62,7 @@ var sigterm_received: std.atomic.Value(bool) = std.atomic.Value(bool).init(false
 const Client = struct {
     alloc: std.mem.Allocator,
     socket_fd: i32,
+    initialized: bool = false,
     has_pending_output: bool = false,
     read_buf: ipc.SocketBuffer,
     write_buf: std.ArrayList(u8),
@@ -139,7 +140,6 @@ const Daemon = struct {
     command: ?[]const []const u8 = null,
     cwd: []const u8 = "",
     has_pty_output: bool = false,
-    has_had_client: bool = false,
     created_at: u64, // unix timestamp (ns)
     is_task_mode: bool = false, // flag for when session is run as a task
     task_exit_code: ?u8 = null, // null = running or n/a, set when task completes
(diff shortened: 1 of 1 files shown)


laudney:
There are some bugs on this; I'm on it.

laudney:
Updated the branch. The main new change in this revision is serializing only the active screen region instead of the full scrollback.

Without this, if the session had previously cleared the screen (e.g. `clear` or Ctrl-L), the snapshot still sent the entire scrollback history. The attaching client would render all that old content — the clear-screen that happened in the session had no effect on the snapshot. The result is the client sees a completely different view from the session, totally out of sync.

Now only the active screen region is serialized, so the client sees exactly what the session displays.

neurosnap:
Hmm, so I think the race condition makes total sense, but I'm less sure about only serializing the active screen. Doesn't that mean we will never have scrollback history beyond the first page? That doesn't sound ideal.

laudney:
agreed; finding a fix...

neurosnap:
I wonder if it would be useful to split this PR out into separate fixes?  I want to see the race condition fix as a separate PR so I can properly test it in isolation.  Thoughts?

laudney:
That was my thought too. Then I found it quite difficult to do all the necessary ablations. It seems to me several issues are inter-linked. I'm being lazy here.

neurosnap:
Or the race condition is not the core issue?  I've done multiple rounds of trying to track down some of these harder-to-understand bugs like the middle cursor one.  I just keep going in circles trying to figure it out all.  Sometimes it's helpful to completely disable state rehydration to see if the underlying impl is correct.

neurosnap:
Checking in, how's it going?  I unfortunately moved some files around so you'll have some conflicts.  Happy to help where I can!

neurosnap:
Closing for now to keep the PR list clean.  Please feel free to reach out if you want to continue working on this PR and I'll reopen.

PR #62, opened 2026-02-25
Add starship shell prompt instructions

ida5428:
Added instructions for using starship as the shell prompt.

### DIFF
--- a/README.md
+++ b/README.md
@@ -154,6 +154,24 @@ POWERLEVEL9K_RIGHT_PROMPT_ELEMENTS+=my_zmx_session
    style = 'plain'
 ```
 
+### starship
+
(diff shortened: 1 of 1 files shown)


injust:
@ida5428 For displaying environment variables, the `env_var` module is more suitable than the custom module.

For example, this is what I use:

```toml
[env_var.ZMX_SESSION]
symbol = " "
format = "[$symbol$env_value]($style) "
description = "zmx session name"
style = "bold sapphire"
```

ida5428:
I didn't have a strong reason to prefer the custom module at the time, It just seemed like a reasonable approach when I wrote it. Looking at it again, using `env_var` makes sense too, and is the simpler option.

PR #74, opened 2026-03-04
fix(run): always single-quote args to avoid bash history expansion bug

### LINKED ISSUE #72: `zmx run` strips quotes, breaking commands with shell metacharacters
When passing a command with quotes through `zmx run`, the quoting is lost before it reaches the session's shell. For example:

```bash
$ zmx run job bash -c 'for i in 1 2 3; do echo "step $i"; sleep 1; done'
session "job" created
command sent

$ zmx history job
bash -c for i in 1 2 3; do echo "step $i"; sleep 1; done ; echo ZMX_TASK_COMPLETED:$?
<...>$ bash -c for i in 1 2 3; do echo "step $i"; sleep
 1; done ; echo ZMX_TASK_COMPLETED:$?
-bash: syntax error near unexpected token `do'
```

The single quotes around the `bash -c` argument are stripped, so the session's shell sees `for`, `do`, `;` etc. as separate tokens instead of a single argument to `bash -c`.

Even simple cases fail:

```bash
$ zmx run job2 uv run python -c "print('hello')"
# session shell sees: uv run python -c print('hello')
# -bash: syntax error near unexpected token `('
```

Workaround: Write the command to a script file and run that instead, `zmx run job /path/to/script.sh`, works fine.

Environment:
- zmx 0.4.1
- Linux x86_64 (Ubuntu 24.04, AWS EC2)
- bash 5.2

kaifronsdal:
## Summary

The re-quoting fix in 954f8db partially solved #72 by adding `shellNeedsQuoting` / `shellQuote` to reconstruct quotes around args containing shell metacharacters. However, `shellQuote` preferred **double quotes**, which breaks `!` (history expansion) in interactive bash:

```
$ zmx run job echo 'hello!'
# shellQuote produced: "hello\!"
# \! does NOT suppress history expansion inside double quotes in interactive bash
```

This PR switches `shellQuote` to **always use single quotes**, matching Python's `shlex.quote`. Inside single quotes nothing is special except `'` itself, which is handled via the standard `'\''` trick (end quote, backslash-escaped literal quote, reopen quote). This is universally safe across bash, zsh, fish, and other POSIX shells.

Also adds comprehensive unit tests for `shellNeedsQuoting` and `shellQuote`.

Fixes #72

## Test plan

- [x] `zig build check` passes
- [x] `zig build test` passes (all existing + new unit tests)
- [x] `zmx run job bash -c 'for i in 1 2 3; do echo "step $i"; sleep 1; done'` -- prints step 1/2/3
- [x] `zmx run job2 python3 -c "print('hello')"` -- prints `hello`
- [x] `zmx run job3 echo 'hello!'` -- prints `hello!` (no history expansion error)
- [x] `` zmx run test echo '`whoami`' `` -- prints literal `` `whoami` `` (backticks not expanded)
- [x] `zmx run test echo '$(whoami)'` -- prints literal `$(whoami)` (not expanded)
- [x] `zmx run test echo '; echo INJECTED'` -- prints literal `; echo INJECTED` (no injection)
- [x] `zmx run test echo '*.zig'` -- prints literal `*.zig` (glob not expanded)
- [x] Both quote types in one arg: `it's "cool"` round-trips correctly

### DIFF
--- a/src/main.zig
+++ b/src/main.zig
@@ -281,25 +281,29 @@ const Daemon = struct {
         const cur_cmd = self.command orelse self.task_command;
         if (cur_cmd) |args| {
             for (args, 0..) |arg, i| {
-                if (i > 0) {
-                    if (cmd_len < ipc.MAX_CMD_LEN) {
-                        cmd_buf[cmd_len] = ' ';
-                        cmd_len += 1;
+                const quoted = if (shellNeedsQuoting(arg))
+                    shellQuote(self.alloc, arg) catch null
+                else
+                    null;
+                defer if (quoted) |q| self.alloc.free(q);
+                const src = quoted orelse arg;
+
+                const need = src.len + @as(usize, if (i > 0) 1 else 0);
+                if (cmd_len + need > ipc.MAX_CMD_LEN) {
+                    const ellipsis = "...";
+                    if (cmd_len + ellipsis.len <= ipc.MAX_CMD_LEN) {
+                        @memcpy(cmd_buf[cmd_len..][0..ellipsis.len], ellipsis);
+                        cmd_len += ellipsis.len;
                     }
+                    break;
                 }
-                if (shellNeedsQuoting(arg)) {
-                    const quoted = shellQuote(self.alloc, arg) catch arg;
-                    defer if (quoted.ptr != arg.ptr) self.alloc.free(quoted);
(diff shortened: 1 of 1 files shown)


neurosnap:
nice job, thanks!

PR #78, opened 2026-03-07
Use bash-completion lazy loading

Self-Perfection:
Use bash-completion lazy loading to avoid launching zmx on each new shell session startup.
Loads completion on demand.

---------

I am sure similar optimizations possible for fish and zsh as well, but I am not familiar with this shells enough to propose them.

### DIFF
--- a/README.md
+++ b/README.md
@@ -183,11 +183,18 @@ ssh remote-server zmx attach session-na<TAB>
 
 ### bash
 
-Add this to your `.bashrc` file:
+Run this snippet:
 
 ```bash
-if command -v zmx &> /dev/null; then
-  eval "$(zmx completions bash)"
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -183,11 +183,18 @@ ssh remote-server zmx attach session-na<TAB>
 
 ### bash
 
-Add this to your `.bashrc` file:
+Run this snippet:
```
neurosnap: I'm curious, why did you make this copy change?

neurosnap (commented):
Hmm, this feels more complicated.  I'm not opposed but I wonder if we should prefer simplicity over performance?  There's also an issue of feature parity with the other shells.

Maybe we should make this an additional section within bash.

neurosnap:
Thanks for the contribution but I'd prefer to keep the shell integrations simple so im gonna close

PR #82, opened 2026-03-09
More bugfixes

chronologos:
Follow-up to #83. 8 more fixes — mostly defensive (the code is demonstrably wrong but you won't hit it in normal use). One actual improvement to `wait`.

There's also a follow-up PR with a Go e2e test harness that exercises a few of these.

---

### `wait` times out instead of polling forever (`src/main.zig:720`)

#83 added the `total > 0` guard so `wait` wouldn't exit 0 when nothing matches. But that just made it poll forever on a typo. Now after ~3s with no matches it exits 2 with "no matching sessions found". 3s is plenty since `run foo && wait foo` is sequential.

---

### Wrong O_NONBLOCK constants + the bugs that hid behind them (`src/main.zig:336,...`)

`0o4000` is the Linux value — on macOS that's `O_EXCL`, which `F_SETFL` just drops. Same for `posix.SOCK.NONBLOCK` (Zig defines it as `0x10000` on darwin for API consistency, but it's not a file-status flag). So on macOS the PTY/socket/stdin were never actually non-blocking. Nothing visibly broke because `poll()` guards every read, but it meant the `WouldBlock` paths were dead code on one of our two platforms.

Verified via `sys/fcntl.h` on macOS 26.3.1 + a C test program that shows `F_SETFL` silently drops both bogus bits.

Fixing this exposed three things that were only working by accident:
- `handleInput`/`handleRun` did `_ = try posix.write(...)` — short writes dropped bytes, `WouldBlock` crashed the daemon. New `ptyWrite()` retries short writes and logs+drops on `WouldBlock` (matches old behavior, just explicit).
- `ensureSession` had both an `errdefer` and a `defer` closing the same fd → double-close → `EBADF` unreachable in safe builds when daemonLoop errored.
- `daemonLoop` error skipped the `handleKill`/`waitpid` that sits below it → orphaned shell. Moved those into the defer.

Also: stdin's O_NONBLOCK lives on the open file description (shared with the parent shell), so we save/restore flags.

---

### SIGPIPE + EINTR handling (`src/main.zig:1349,1223`)

SIGPIPE's default is terminate, and it fires *before* `write()` gets to return `BrokenPipe`. So an abruptly exiting client could kill the daemon. `ignoreSigpipe()` in `main()` fixes it (inherited across fork). The SIG_IGN is restored to SIG_DFL before exec so `yes | head` inside a session still exits 141.

Daemon's poll loop didn't handle EINTR — SIGTERM made it error out instead of checking the flag and exiting cleanly. Fixed.

Deliberately *not* setting `SA_RESTART` on the handlers: on BSD/macOS `poll()` is restartable with that flag, so an idle daemon would never wake up to check the sigterm flag.

---

### Stale-socket cleanup on Timeout orphans busy daemons (`src/util.zig:46`, ...)

The 1s probe timeout can fire on a daemon that's just busy (heavy output, resize reflow). Deleting its socket orphans it forever. Now only `ConnectionRefused` triggers cleanup. `zmx list` now says `status=unreachable` instead of the misleading `cleaning up`. `zmx wait` treats unreachable sessions as done+failed so it doesn't loop forever.

---

### Fork-child can fall through to parent code path (`src/main.zig:303`)

`try` in the `pid == 0` branch could propagate an error past the if-block — the child then runs the parent code, maybe starting a second daemon or deleting the real one's socket via errdefer. The `bufPrintZ` case triggers with a `SHELL` basename >62 chars. There was also a fixed `argv_buf[64]` with no bounds check.

Extracted an `execChild()` returning `!noreturn` — it either execs or returns an error; caller exits on any error. Argv is heap-allocated now.

---

### `handleInit` serialize/resize order (`src/main.zig:426`)

Comment says "serialize BEFORE resize", code did the opposite. Resize triggers reflow which moves the cursor, so the snapshot had a stale cursor position. Swapped the order.

---

### Undefined termios with non-TTY stdin (`src/main.zig:860`)

`_ = tcgetattr(...)` discarded the error. With piped stdin it fails and `orig_termios` is stack garbage. Fixed by checking the return and skipping termios setup when stdin isn't a TTY.

---

### IPC wire data validation (`src/ipc.zig:53`, ...)

Stuff coming off the socket wasn't sanity-checked:
- `header.len` + `@sizeOf(Header)` could wrap on a near-u32-max value → widened to usize
- `cmd_len`/`cwd_len` used directly as slice bounds → clamped with `@min`
- `Tag` enum was exhaustive, so an out-of-range byte was UB → made non-exhaustive with `_` arm
- Same for `HistoryFormat` → `intToEnum catch .plain`

### DIFF
--- a/src/ipc.zig
+++ b/src/ipc.zig
@@ -15,6 +15,10 @@ pub const Tag = enum(u8) {
     History = 8,
     Run = 9,
     Ack = 10,
+    // Non-exhaustive: this enum comes off the wire via bytesToValue and
+    // @enumFromInt, so out-of-range values (11-255) are representable
+    // rather than UB. Switches must handle `_` (unknown tag).
+    _,
 };
 
 pub const Header = packed struct {
@@ -53,7 +57,9 @@ pub const Info = extern struct {
 pub fn expectedLength(data: []const u8) ?usize {
     if (data.len < @sizeOf(Header)) return null;
     const header = std.mem.bytesToValue(Header, data[0..@sizeOf(Header)]);
-    return @sizeOf(Header) + header.len;
+    // header.len comes off the wire; widen to usize before adding so a
+    // near-u32-max value can't wrap (panic in safe mode, UB in release).
+    return @as(usize, @sizeOf(Header)) + @as(usize, header.len);
 }
 
 pub fn send(fd: i32, tag: Tag, data: []const u8) !void {
--- a/src/main.zig
+++ b/src/main.zig
@@ -652,6 +652,7 @@ fn wait(cfg: *Cfg, session_names: std.ArrayList([]const u8)) !void {
     // appeared yet" (keep polling) from "sessions we were tracking
     // disappeared" (fail — daemon crashed or was killed).
     var max_seen: i32 = 0;
+    var zero_match_iters: u32 = 0;
 
     while (true) {
         var sessions = try util.get_session_entries(alloc, cfg.socket_dir);
@@ -704,6 +705,20 @@ fn wait(cfg: *Cfg, session_names: std.ArrayList([]const u8)) !void {
             try stdout.flush();
             std.process.exit(agg_exit_code);
             return;
+        }
+
+        if (max_seen == 0) {
+            // `zmx run foo && zmx wait foo` is essentially sequential, so
+            // matching sessions should be visible from the first poll. If
+            // nothing appears after a few iterations it's almost certainly a
+            // typo, not a slow start.
+            zero_match_iters += 1;
+            if (zero_match_iters >= 3) {
+                try stdout.print("error: no matching sessions found\n", .{});
+                try stdout.flush();
+                std.process.exit(2);
+                return;
+            }
         }
 
         std.Thread.sleep(1000 * std.time.ns_per_ms);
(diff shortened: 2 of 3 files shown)


neurosnap:
Hi! Thank you so much for these contributions, they were excellent.  I checked out every commit individually and made minor tweaks but otherwise really nice job.  I'm curious, how did you discover these bugs?  Did you have a code agent run through some code review or do you have experience with socket programming?

The one thing I'd like to follow-up on is making stdin writes go through `poll()` instead of the best effort solution you wrote.

The other issue I've been wrestling with is when the ipc payloads change and the daemon can no longer communicate with zmx.  I'm not sure if you have any suggestions there but I'd be happy to work through it with you.

Your changes were pushed to `main` so I'm closing this PR.  Thanks again!

neurosnap:
In case you want to test more:

- https://zmx.sh/a/zmx-0.4.2-pre-linux-aarch64.tar.gz
- https://zmx.sh/a/zmx-0.4.2-pre-linux-x86_64.tar.gz
- https://zmx.sh/a/zmx-0.4.2-pre-macos-aarch64.tar.gz
- https://zmx.sh/a/zmx-0.4.2-pre-macos-x86_64.tar.gz

chronologos:
Thanks! I have some experience in socket programming in C and used Claude to identify and triage a list of potential bugs, and worked through each bug with it as well. I'll think a bit more about ipc payload changes and making stdin writes go through poll() too

PR #83, opened 2026-03-09
Fix verified bugs

chronologos:
5 fixes for user-observable bugs, all A/B-verified on macOS. Each commit builds on its own (bisect-safe).

All five scenarios fail on `main`, pass on this branch:

| # | Scenario | `main` | This branch |
|---|----------|--------|-------------|
| 1 | `zmx attach "../evil"` with `ZMX_DIR=/tmp/test` | Socket created at `/tmp/evil` (outside socket dir) | `error.InvalidSessionName` |
| 2 | `zmx history --vt` (no session arg) | Panic on `.?` unwrap of null | `error.SessionNameRequired` |
| 3 | `zmx wait <nonexistent>` | Exit 0 immediately | Keeps polling |
| 4 | 2nd `zmx run sess "false"` after `zmx run sess "true"` → `zmx wait sess` | Exit 0 (stale, and 2nd cmd never executed) | Exit 1 |
| 5 | 3× run/wait: true, false, true | `0 / 0 / 0` | `0 / 1 / 0` |

### 1. Path traversal via session name (`src/socket.zig:8`)

Session names become filenames under `socket_dir`, but `getSeshName` did no validation. `zmx attach "../foo"` created a socket outside that directory; stale-socket cleanup via `unlinkat(dirfd, "../foo")` deleted files outside it. `ZMX_SESSION_PREFIX` was similarly unsanitized.

**Fix:** Reject names containing `/`, NUL, `.`, or `..`. Also fixed a minor double-call of `seshPrefix()`.

---

### 2. `zmx history` panics on missing session name (`src/main.zig:86`)

`zmx history --vt` (flags only, no positional) left `session_name` null; `session_name.?` panicked.

**Fix:** `session_name orelse ""` so `getSeshName` handles it (returns `SessionNameRequired` if no `ZMX_SESSION_PREFIX` either).

---

### 3. `zmx wait` vacuous success & disappeared-session loop (`src/main.zig:679`)

`if (total == done)` was true when both were 0. A typo in the session name, or `wait` polling before the preceding `zmx run` created its socket → exit 0 immediately.

**Fix:** Require `total > 0`. Also tracks a high-water mark (`max_seen`) across iterations: if the match count drops, a session disappeared (daemon crashed/killed) → exit 1 with an error message instead of silently looping forever. The disappearance check runs *before* the completion check so a crashed session doesn't get masked by others completing.

---

### 4. Stale `task_exit_code` — subsequent `zmx run` ignored (`src/main.zig:589`)

The daemon's exit-marker scan is gated on `task_exit_code == null`, but that field was never reset after the first task completed. A second `zmx run` on the same session had its marker ignored; `zmx wait` returned the *first* task's exit code.

**Fix:** Reset `task_exit_code`, `task_ended_at`, and set `is_task_mode=true` in `handleRun`.

---

### 5. `zmx run` sends `\n`, but readline wants `\r` (`src/main.zig:950`)

When the shell is at the readline prompt (raw mode), `\n` is just cursor-down — readline binds accept-line to `\r`. The first-ever `zmx run` on a fresh session worked only because it arrives during shell startup while the line discipline is still canonical (ICRNL translates `\n`). Any subsequent `run` left the command typed but unexecuted at the prompt.

**Bugs #4 and #5 masked each other:** unpatched, the second command never executes, but `wait` returns the first's code anyway — looks fine until you check `$?`.

**Fix:** Use `\r`. Works in both modes — canonical mode translates it via ICRNL, raw-mode readline accepts it directly. Applied to both inline-arg and stdin-piped paths.

### DIFF
--- a/src/main.zig
+++ b/src/main.zig
@@ -970,7 +970,11 @@ fn run(daemon: *Daemon, command_args: [][]const u8) !void {
         }
 
         try cmd_list.appendSlice(alloc, inline_task_marker);
-        try cmd_list.append(alloc, '\n');
+        // \r, not \n: once the shell is at the readline prompt the PTY is in
+        // raw mode; readline's accept-line binds to CR. The first-ever run
+        // works with \n only because it arrives during shell startup while
+        // the line discipline is still canonical.
+        try cmd_list.append(alloc, '\r');
 
         cmd_to_send = try cmd_list.toOwnedSlice(alloc);
         allocated_cmd = @constCast(cmd_to_send.?);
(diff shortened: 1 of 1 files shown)


neurosnap (approved):
These contributions are great, thanks so much!  I'll merge and continue to QA on main

PR #85, opened 2026-03-11
fix: validate session name length against Unix socket path limit

### LINKED ISSUE #84: Session names exceeding Unix socket path limit fail silently
### Description

When a session name causes the full Unix socket path to exceed the OS `sun_path` limit (104 bytes on macOS, 108 on Linux), zmx commands fail silently — exit code 1 with no error output. There is no validation on session name length and no error message indicating why the operation failed.

### Root Cause

zmx constructs socket paths as `<socket_dir>/<session_name>`. The socket directory can be quite long (e.g., 56 bytes on macOS with `$TMPDIR/zmx-<uid>`), leaving very few bytes for the session name. When the total path hits the `sun_path` limit (minus null terminator), `initUnix` fails and the error is swallowed.

### Reproduction

```bash
# Check your socket dir length
# macOS with default TMPDIR: often ~56 bytes, leaving only 46 chars for session name

# Works (46 chars):
zmx run "my-very-important-project-feature-branch-ses" echo hello
# => session "my-very-important-project-feature-branch-ses" created

# Fails silently (53 chars):
zmx run "my-very-important-project-feature-branch-session-name" echo hello
# => (no output, exit code 1)
```

All commands are affected: `run`, `attach`, `kill`, `history`. Some return exit code 1 silently (`run`, `attach`), others return exit code 0 despite doing nothing (`kill`, `history`).

### Environment

- zmx 0.3.0
- macOS (Darwin 25.2.0, `sun_path` = 104 bytes, confirmed via compiled C)
- Linux (`sun_path` = 108 bytes, confirmed via Docker with both compiled C and Python socket bind test)
- Socket dir: `$TMPDIR/zmx-<uid>` (56 bytes on test machine)
- Max working session name on test machine: **46 characters**

### Expected Behavior

- zmx should validate the session name length before attempting socket operations
- On failure, zmx should print a clear error message, e.g.: `error: session name too long (max 46 characters for current socket directory)`
- Ideally, compute the limit dynamically based on the active socket directory: `103 - len(socket_dir) - 1` (macOS) or `107 - len(socket_dir) - 1` (Linux)

### Additional Context

The max session name length varies by platform and configuration:

| Platform | `sun_path` size | Usable path limit | Typical socket dir | Max session name |
|---|---|---|---|---|
| macOS (TMPDIR) | 104 | 103 | ~56 bytes | ~46 chars |
| Linux (XDG_RUNTIME_DIR) | 108 | 107 | ~21 bytes | ~85 chars |

mdsakalu:
## Summary

Session names that cause the full socket path to exceed the OS `sun_path` limit (104 bytes on macOS, 108 on Linux) now produce a clear error message instead of failing silently.

**Before:** `zmx run "my-very-important-project-feature-branch-session-name" echo hello` → no output, exit code 1
**After:** `error: session name is too long (53 bytes, max 46 for socket directory "/var/folders/.../zmx-502")`

- Add `max_socket_path_len` comptime constant derived from the platform's `sockaddr_un.path` field (no hardcoded values)
- Validate in `getSocketPath` before allocating, returning `error.NameTooLong`
- Add `maxSessionNameLen` helper to compute the dynamic limit for a given socket directory
- Print a clear error to stderr in all command paths: `run`, `attach`, `kill`, `history`, `detach`, `list`
- Move validation before `sessionExists` in `kill`/`history` so the error is shown even when the socket file doesn't exist
- Add 6 unit tests covering boundary conditions and platform constants

Fixes #84

## Test plan

- [x] Unit tests pass (`zig build test`)
- [x] 46-char session name succeeds (at limit on macOS with default TMPDIR)
- [x] 53-char session name shows clear error on `run`, `attach`, `kill`, `history`
- [x] Verified `max_socket_path_len` is 103 on macOS, 107 on Linux (via Docker)

### DIFF
--- a/src/main.zig
+++ b/src/main.zig
@@ -127,7 +127,10 @@ pub fn main() !void {
             .cwd = cwd,
             .created_at = @intCast(std.time.timestamp()),
         };
-        daemon.socket_path = try socket.getSocketPath(alloc, cfg.socket_dir, sesh);
+        daemon.socket_path = socket.getSocketPath(alloc, cfg.socket_dir, sesh) catch |err| switch (err) {
+            error.NameTooLong => return printSessionNameTooLong(sesh, &cfg),
+            error.OutOfMemory => return err,
+        };
         std.log.info("socket path={s}", .{daemon.socket_path});
         return attach(&daemon);
     } else if (std.mem.eql(u8, cmd, "run") or std.mem.eql(u8, cmd, "r")) {
@@ -159,7 +162,10 @@ pub fn main() !void {
             .is_task_mode = true,
             .task_command = cmd_args_raw.items,
         };
-        daemon.socket_path = try socket.getSocketPath(alloc, cfg.socket_dir, sesh);
+        daemon.socket_path = socket.getSocketPath(alloc, cfg.socket_dir, sesh) catch |err| switch (err) {
+            error.NameTooLong => return printSessionNameTooLong(sesh, &cfg),
+            error.OutOfMemory => return err,
+        };
         std.log.info("socket path={s}", .{daemon.socket_path});
         return run(&daemon, cmd_args_raw.items);
     } else if (std.mem.eql(u8, cmd, "wait") or std.mem.eql(u8, cmd, "w")) {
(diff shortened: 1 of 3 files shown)


neurosnap:
Thanks! I pulled your PR and move some things around

PR #87, opened 2026-03-11
feat: add session rename command

### LINKED ISSUE #46: Feature request: rename sessions
It would be nice to be able to rename existing sessions.

For example: I often start sessions for short tasks with useless throwaway names, but sometimes the sessions end up becoming longer-lived, so being able to reassign them to a more descriptive name would be desirable. This is definitely a me problem, but this is one use case where this feature would be very useful!

mdsakalu:
## Summary

Adds `zmx rename <old> <new>` (alias `rn`) to rename a running session.

Closes #46

- New `Rename` IPC tag (11) for client-to-daemon communication
- Daemon renames socket file + log file atomically, updates internal state, broadcasts notification to attached clients
- Client-side `rename()` follows the `kill()` pattern (probe, send, wait for ack)
- `zmx rename <new>` shorthand works inside a session (reads `$ZMX_SESSION`)
- Shell wrapper functions (bash/zsh/fish) update `$ZMX_SESSION` after successful in-session rename

## Known Limitations

See discussion on #46 — the `$ZMX_SESSION` environment variable cannot be updated in a running shell by an external process (fundamental Unix constraint). The shell wrapper in the completion scripts handles the in-session case, but when another process renames your session, attached clients receive a terminal notification only.

## Test plan

- [ ] `zmx attach test` then `zmx rename test new-test` from another terminal — notification shown, `zmx ls` shows new name
- [ ] Inside a session: `zmx rename new-name` — works, `echo $ZMX_SESSION` shows new name (with shell wrapper sourced)
- [ ] Error cases: nonexistent session, target name already exists, invalid characters
- [ ] Session continuity after rename: typing, output, detach/reattach via new name
- [ ] `zmx kill <new-name>` works after rename
- [ ] Log file renamed under `{socket_dir}/logs/`

**Note:** Build cannot be verified right now — ghostty dependency CDN is returning 403s (affects trunk too).

### DIFF
--- a/src/completions.zig
+++ b/src/completions.zig
@@ -23,13 +23,23 @@ pub const Shell = enum {
 };
 
 const bash_completions =
+    \\zmx() {
+    \\  command zmx "$@"
+    \\  local rc=$?
+    \\  if [[ $rc -eq 0 && ("$1" == "rename" || "$1" == "rn") && -n "$ZMX_SESSION" ]]; then
+    \\    local new_name="${3:-$2}"
+    \\    [[ -n "$new_name" ]] && export ZMX_SESSION="$new_name"
+    \\  fi
+    \\  return $rc
+    \\}
+    \\
     \\_zmx_completions() {
     \\  local cur prev words cword
     \\  COMPREPLY=()
     \\  cur="${COMP_WORDS[COMP_CWORD]}"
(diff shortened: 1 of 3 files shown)


neurosnap:
I'm going to close this PR for now since we are still actively discussing solutions.  When we are ready with a solution we can revisit.  Sorry for any inconvenience and thank you very much for this contribution.

PR #92, opened 2026-03-15
fix(attach): handle all kitty keyboard protocol encodings for Ctrl+\

sundbp:
The old detection used rigid substring matching for exactly two sequences (\x1b[92;5u and \x1b[92;5:1u). This failed when apps enabled progressive enhancement flags that change the encoding:

- Lock modifiers (caps_lock, num_lock) alter the modifier value
- Alternate key sub-fields add :shifted:base after the key code
- Combined modifiers (ctrl+shift, ctrl+alt, etc.) change the value
- Text codepoint sections append an extra ;codepoints field

Replace with a proper CSI u parser that:
1. Matches key code 92 (backslash) with any alternate sub-fields
2. Checks the ctrl bit in the modifier value: (mod - 1) & 0b100
3. Accepts press/repeat events, rejects release
4. Tolerates optional text codepoint sections

### DIFF
--- a/src/util.zig
+++ b/src/util.zig
@@ -246,8 +246,11 @@ fn parseKittyCtrlBackslash(buf: []const u8) bool {
     if (mod_encoded < 1) return false;
     const mod_raw = mod_encoded - 1;
 
-    // 5. Ctrl bit (0b100 = 4) must be set.
-    if (mod_raw & 0b100 == 0) return false;
+    // 5. Ctrl must be the only intentional modifier. Lock modifiers
+    //    (caps_lock=0b1000000, num_lock=0b10000000) are tolerated because
+    //    they are ambient state, not deliberate key combinations.
+    const intentional_mods = mod_raw & 0b00111111;
+    if (intentional_mods != 0b100) return false;
 
     // 6. Parse optional event type after ':'.
     if (pos < buf.len and buf[pos] == ':') {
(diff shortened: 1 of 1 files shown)


neurosnap:
Hi, thanks for submitting this PR.  The reason why I originally went with a simpler check was because of the complexity involved with covering all these cases.  Having said that, happy to talk through our options because I do want the detach to "just work."

> Lock modifiers (caps_lock, num_lock) alter the modifier value

I'm honestly not sure caps_lock is in scope, that seems like something the end-user should toggle.  I could potentially see adding support for num_lock since people might just have that on as a habit.

> Alternate key sub-fields add :shifted:base after the key code

Could you explain under what conditions this is relevant?

> Combined modifiers (ctrl+shift, ctrl+alt, etc.) change the value

I'm not convinced this is in scope, why would we support the user pressing `ctrl+shift/alt+\`?

> Text codepoint sections append an extra ;codepoints field

Could you explain under what conditions this is relevant?

sundbp:
Hi. I started on this after realizing that I couldn't detach in certain applications on one of my laptops or using a 75% keyboard of mine (without numlock/scrollock etc). The common cause seemed to be kitty protocol apps.

On the modifiers I think this one is very valid. And the main culprit for my specific use case. The spec is clear: when a program inside the session enables the "Report all keys as escape codes" flag (0b1000), the terminal will include lock modifier bits in the encoded value. Programs like neovim, helix, and kakoune enable this flag. If a user has num_lock on (extremely common -- many people leave it on permanently, or as in my 75% case it may be set to on without any way to physically turning it off) and is running neovim inside zmx, the terminal sends \x1b[92;133u (133 = 1 + 4 + 128) instead of \x1b[92;5u. Without handling this, detach silently breaks.

Regarding sub-fields:
When the inner program enables "Report alternate keys" (flag 0b100), the terminal inserts :shifted_key:base_layout_key sub-fields after the key code. For \, the shifted key is | (124), producing \x1b[92:124;5u. The parser needs to skip these to find the ; separator. Programs routinely enable multiple flags at once (e.g., neovim uses 0b11111), so this is a real encoding that appears in practice.

on the extra modifiers along with Ctrl: You're right - this is not sensible and I'll rever this part.

Regarding codepoints:
When "Report associated text" (flag 0b10000) is enabled, the terminal appends ;28 (the text codepoint for ctrl+) before the terminal u. Without skipping this field, the parser fails to match. This flag is typically combined with "Report all keys" (0b1000), so it appears whenever a program requests the full kitty keyboard protocol (flags=0b11111).

neurosnap:
Merged with some modifications, thanks!

PR #93, opened 2026-03-16
feat: add optional TCP transport for cross-boundary session access

schuyler:
## Summary

- Add `--tcp addr:port` flag to `zmx attach` that creates a TCP listener alongside the Unix socket, enabling clients to connect from containers, VMs, or remote hosts
- `zmx attach addr:port` and `zmx run addr:port` connect to sessions via TCP when the argument looks like a `host:port` address
- Supports numeric IPs (v4/v6), hostnames (e.g. `host.docker.internal`), and bare ports (`:7777` defaults to `0.0.0.0`)
- TCP_NODELAY set on all TCP connections for interactive responsiveness
- Prints a warning when binding to a non-loopback address (no authentication in this version)

## Example usage

```bash
# Host: create a session with TCP listener
zmx attach --tcp 0.0.0.0:7777 my-session

# Container: connect via TCP
zmx attach host.docker.internal:7777
zmx run host.docker.internal:7777 echo hello
```

## Test plan

- [x] `zig build` compiles
- [x] `zig build test` passes (new tests for address parsing)
- [x] Local smoke test: create TCP session, attach, run, kill
- [x] Cross-boundary test: Docker container on Colima connects to host session via `host.docker.internal`

### DIFF
--- a/src/main.zig
+++ b/src/main.zig
@@ -443,7 +443,7 @@ const Daemon = struct {
 
             // Create TCP listener if --tcp was specified
             if (self.cfg.tcp_addr) |tcp_addr| {
-                const tcp_fd = socket.createTcpSocket(tcp_addr) catch |err| {
+                const tcp_fd = socket.createTcpSocket(self.alloc, tcp_addr) catch |err| {
                     posix.close(server_sock_fd);
                     dir.deleteFile(self.session_name) catch {};
                     std.log.err("failed to create TCP listener on {s}: {s}", .{ tcp_addr, @errorName(err) });
@@ -1127,7 +1127,7 @@ fn attachTcp(addr: []const u8) !void {
         return error.CannotAttachToSessionInSession;
     }
 
-    const client_sock = try socket.tcpConnect(addr);
+    const client_sock = try socket.tcpConnect(std.heap.c_allocator, addr);
     std.log.info("attached via tcp={s}", .{addr});
 
     var orig_termios: cross.c.termios = undefined;
@@ -1237,7 +1237,7 @@ fn runTcp(alloc: std.mem.Allocator, addr: []const u8, command_args: [][]const u8
     }
 
     // Connect via TCP and probe
-    const fd = try socket.tcpConnect(addr);
+    const fd = try socket.tcpConnect(alloc, addr);
     errdefer posix.close(fd);
 
     const info = ipc.probeSessionFd(alloc, fd) catch |err| {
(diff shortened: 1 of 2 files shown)


neurosnap:
Hi, thanks for submitting this PR.  I have to say that I'm intrigued. I see in your example that you connecting to a session through docker but I'm wondering more specifically what your use case is for zmx?

schuyler:
> Hi, thanks for submitting this PR. I have to say that I'm intrigued. I see in your example that you connecting to a session through docker but I'm wondering more specifically what your use case is for zmx?

I sometimes run interactive terminal sessions through zmx on my local system, so that I can watch what Claude Code is doing and possibly intervene. I have a custom Claude Code command called `/zmx-session` that I've attached to this ticket if you're curious.

[SKILL.md](https://github.com/user-attachments/files/26062286/SKILL.md)

At the same time, I also run Claude Code in [packnplay](https://github.com/obra/packnplay), so that it's isolated from my local system. I'd like to have Claude Code be able to interact with zmx sessions running on the host machine or in other containers.

I looked at using `socat` or SSH UNIX socket forwarding, but those solutions were all pretty hacky compared to just leveraging zmx's existing transport abstraction.

Does that make sense? I realize it's a bit of a leap for zmx and maybe not a direction you want to go in.

neurosnap:
That's really cool! I like the skill, we should probably commit that to the repo inside a `contrib/` folder.  I was idly working on a prototype code agent that used `zmx` for its context so there wouldn't be a TUI and instead of asking the agent to run commands you could just run the commands and the agent would see the results in your pty.

As for this PR specifically, while really neat, feels a little out-of-scope.  However, I'd love to have these types of contributions checked into our repo somehow.  I'm curious, would you be at all interested in using this feature by applying a patch on top of `zmx`?  Sort of like how `dwm` works?  We would check in a set of patches inside of a `patches/` folder and then users could clone the repo, apply the patches they want, and then build their own `zmx` binary?

If you have other ideas on how we could support extensions like this one I'd be happy to brainstorm with you.

schuyler:
Please feel free to re-use the skill!

The only thing I can think to suggest is adding a transport plugin architecture to zmx, but that also might not jive with your design. I appreciate the patch directory option, but that seems potentially brittle and tricky to maintain.

The alternative is I go back to the drawing board and come up with the least hacky way to share UNIX sockets between containers, and go with that. I could write a recipe or howto doc and submit it as a documentation patch against this repo, if you'd like. I have a hunch that my use case for your tool is fringe enough but not too fringe.

Regardless of what you wind up deciding, I'd like to say thanks for zmx -- it fills a useful niche for which most of the other available solutions are overkill. I'm happy to contribute to the project in whatever way is most helpful for your users, so please let me know.

PR #95, opened 2026-03-16
Avoid replaying synchronized output on reattach

figelwump:
## Summary
- stop replaying DECSET 2026 synchronized-output mode during client reattach
- keep other terminal mode restoration intact
- add regression coverage for synchronized-output exclusion during restore

## Why
`synchronized_output` is a transient rendering handshake between a program and its current terminal client. Replaying it into a newly attached client can cause that client to defer renders until its local timeout fires, which makes restored sessions appear sluggish even though the underlying session is healthy.

This mode is different from durable terminal configuration like bracketed paste or wraparound: it should apply to the active client, not become part of long-lived session state.

## Implementation
`serializeTerminalState()` now temporarily clears `term.modes.synchronized_output` before formatting VT state, then restores the in-memory bit before returning. This preserves zmx's existing state model while preventing reattached clients from inheriting a stale synchronized-output state.

## Testing
- `zig build test`
- `zig build`

### DIFF
--- a/src/util.zig
+++ b/src/util.zig
@@ -212,6 +212,17 @@ pub fn serializeTerminalState(alloc: std.mem.Allocator, term: *ghostty_vt.Termin
     var builder: std.Io.Writer.Allocating = .init(alloc);
     defer builder.deinit();
 
+    // Synchronized output (DECSET 2026) is a transient rendering handshake
+    // between a program and its current terminal client. Replaying it to a
+    // newly attached client can leave that client deferring renders until its
+    // local timeout fires, so temporarily exclude it from restored state and
+    // restore the original mode before returning.
+    const had_synchronized_output = term.modes.get(.synchronized_output);
+    if (had_synchronized_output) {
+        term.modes.set(.synchronized_output, false);
+        defer term.modes.set(.synchronized_output, true);
+    }
+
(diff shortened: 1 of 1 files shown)


figelwump:
This fixes an issue I saw in Toastty (https://github.com/figelwump/toastty) where an embedded Ghostty client running zmx was frequently deferring renders in some cases. For ex, running codex within zmx, when you logout and log back in to codex the terminal animation was super slow and janky; ghostty debug logs showed the deferred renders.

figelwump:
Yes, I did test the patch with toastty and it seemed to resolve the deferred renders I was seeing; definitely a noticeable difference in perceived smoothness. 

I think the fix belongs in zmx. Ghostty app does clear synchronized_output on Termio.resize, but zmx reattach doesn’t seem to go through that path. In zmx, the restore payload is serialized and sent first, then TIOCSWINSZ is applied, then the term.resize runs. That lower-level resize doesn’t clear synchronized_output, and it no-ops when the size is unchanged. So the attaching client can still receive a stale ?2026h in the restore blob. To me that makes this something to be fixed in zmx rather than something Toastty should special-case.

I could be wrong though, open to other thoughts. Also, great work with zmx, I have moved over to it from tmux. The next version of toastty will introduce terminal profiles as a first class concept and my main example is using zmx for session persistence.

Issue #96, opened 2026-03-16
Testing framework


neurosnap:
We have tiny unit tests in the codebase but nothing that confirms fundamentally things are working correctly.

I'm not sure what's feasible but there are a lot of user PRs in the mix that are hard to confirm they won't break current behavior without manually testing.

I'd love to have some automated tests that help instill confidence in code changes.

Class of tests:
- Daemon + client interactions (1 client, N client)
- All commands ran with some basic param testing and covering basic functionality
- Basic rehydration functionality given different shells (and terminal emulators?)

Any help or thoughts would be greatly appreciated!

ikma-ricon:
We've been building integration tests for zmx and found two patterns that might be useful here:

**1. Per-test `ZMX_DIR` isolation**

Each test sets `export ZMX_DIR="/tmp/test-$$"` to get its own socket directory. This means:
- No cross-test interference (each test has a clean slate)
- Trivial cleanup (`rm -rf $ZMX_DIR` in teardown)
- No zombie sockets from crashed tests polluting subsequent runs

**2. Short socket paths**

Unix domain sockets are limited to ~104 bytes on macOS. Test frameworks like bats use deeply nested temp dirs that eat most of that budget, causing `zmx run` to silently fail with `session name is too long`. Using a short `ZMX_DIR` path (e.g., `/tmp/sht-$$`) leaves room for reasonable session names.

**3. FD inheritance (PR #107)**

The forked daemon inherits all parent FDs, which causes test harnesses that wait for FD closure (bats, etc.) to hang indefinitely. We submitted #107 to fix this.

Our test suite: [KnickKnackLabs/shell](https://github.com/KnickKnackLabs/shell/tree/main/test) — 22 bats tests exercising run, list, kill, history, status, and wait. Happy to contribute these patterns if you go with bats or a similar framework.

neurosnap:
We merged bats which helps a lot with integration tests. Marking this as closed.

Issue #86, opened 2026-03-11
cursor style error after attach session


GladioFeng:
The bar remained steady throughout the session.
<img width="904" height="117" alt="Image" src="https://github.com/user-attachments/assets/9a0ee60c-a498-4e2c-933a-ce56516f6831" />

Attaching to the session causes it to block.
<img width="794" height="135" alt="Image" src="https://github.com/user-attachments/assets/9afcb394-2bd1-4972-af85-61133764b5d8" />

neurosnap:
Hmm, ghostty must be doing something different when we rehydrate the session.

neurosnap:
#146

neurosnap:
We made some changes that hopefully fix this, closing for now, but if the issue comes back i'll reopen.

GladioFeng:
This issue still persists.  
When I attach the ZMX session, it becomes unresponsive and blocks unexpectedly.  

Here is the fish script I use to attach the ZMX session:  

```fish
function zmx_select
    set -l display (zmx list 2>/dev/null | awk -F'\t' '{
        for (i = 1; i <= NF; i++) sub(/^[[:space:]]+/, "", $i)
    }
    $1 ~ /^name=/ {
        sub(/^name=/, "", $1)
        sub(/^pid=/, "", $2)
        sub(/^clients=/, "", $3)
        sub(/^started_in=/, "", $5)
        sub(/^[[:space:]]+/, "", $1)
        printf "%-20s  pid:%-8s  clients:%-2s  %s\n", $1, $2, $3, $5
    }')

    if test -z "$display"
        echo "⚠️  No zmx session found."
        return 1
    end

    set -l fzf_output (printf "%s\n" $display | fzf \
        --print-query \
        --expect=ctrl-n \
        --height=100% \
        --reverse \
        --prompt="zmx> " \
        --header="Enter: select | Ctrl-N: create new" \
        --preview='zmx history {1}' \
        --preview-window=right:50%:follow | string collect)

    set -l rc $status

    set -l lines (string split \n $fzf_output)
    set -l query $lines[1]
    set -l key $lines[2]
    set -l selected $lines[3]
    set -l session_name ""

    if test "$key" = ctrl-n; and test -n "$query"
        set session_name "$query"
    else if test $rc -eq 0; and test -n "$selected"
        set session_name (echo "$selected" | awk '{print $1}')
    else if test -n "$query"
        set session_name "$query"
    else
        return 130
    end

    set fish_bind_mode insert
[...]

PR #98, opened 2026-03-18
fix: two-phase serialization for nested session cursor corruption

### LINKED ISSUE #31: On reattach the active line is in the middle of the screen

mdsakalu:
## Summary

Fixes #31 — cursor position corruption on re-attach over SSH (nested zmx sessions).
Partially addresses #86 — cursor style error after attach session.
Contributes to #96 — adds 7 integration tests to the testing framework.

## Problem

When running nested zmx sessions (`zmx→SSH→zmx`), re-attaching to the outer session causes visible content to appear at wrong rows. The cursor position and screen content are shifted by the number of scrollback lines that overflow the terminal.

**Root cause:** `serializeTerminalState()` writes scrollback + visible content sequentially. Scrollback lines scroll the terminal before CUP (cursor position) sequences arrive, shifting all visible content down.

This was verified with an automated test suite that detects a consistent +N row shift in the VT bytes sent to the user's terminal during re-attach.

## Fix

Split `serializeTerminalState()` into two phases with a clear screen between:

1. **Phase 1:** Emit scrollback content only (no modes/cursor/keyboard). These lines scroll into the terminal's scrollback buffer — preserved for native scroll-up.
2. **Clear visible screen** (`ESC[2J ESC[H ESC[0m`) — resets the viewport.
3. **Phase 2:** Emit visible screen only with full extras (modes, cursor, keyboard, scrolling region).

The clear between phases ensures visible content starts from a clean slate regardless of how much scrollback preceded it. Inspired by tmux's visible-only redraw approach, but preserves terminal scrollback (tmux doesn't send scrollback to the terminal at all).

**No behavior change for non-nested sessions** — the two-phase approach produces identical visible results when there's no nesting. Scrollback is preserved in the terminal's buffer for both cases.

## Tests

Added 7 integration tests using a ghostty-vt roundtrip pattern:

| Test | What it verifies |
|------|------------------|
| cursor position | Cursor at (10,20) survives serialize→restore |
| CUP-positioned markers | 4 markers at specific rows/cols survive roundtrip |
| scrollback + visible | **80 lines scrollback does not shift markers** (the core bug) |
| nested double-roundtrip | Inner→outer→client double serialization |
| alternate screen | Alt screen content stays on alt screen |
| size mismatch | 30→24 row resize + roundtrip |
| stress (scrollback + size + nesting) | All corruption vectors combined |

All tests run in **<1 second** via `zig build test`. The scrollback and nested tests fail without the fix (scrollback lines appear in the visible area) and pass with it.

## Test plan

- [x] `zig build test` passes (9 tests total: 2 existing + 7 new)
- [x] `zig build` produces working binary
- [x] Verified tests fail without fix (revert → 3 failures with clear error messages)
- [x] Verified tests pass with fix (all 9 pass)
- [ ] Manual testing with real SSH nested sessions
- [ ] Verify scrollback is preserved in terminal after re-attach (non-nested)

### DIFF
--- a/src/util.zig
+++ b/src/util.zig
@@ -223,9 +223,67 @@ pub fn serializeTerminalState(alloc: std.mem.Allocator, term: *ghostty_vt.Termin
         defer term.modes.set(.synchronized_output, true);
     }
 
-    var term_formatter = ghostty_vt.formatter.TerminalFormatter.init(term, .vt);
-    term_formatter.content = .{ .selection = null };
-    term_formatter.extra = .{
+    const pages = &term.screens.active.pages;
+    const screen_top = pages.getTopLeft(.screen);
+    const active_top = pages.getTopLeft(.active);
+    const has_scrollback = !screen_top.eql(active_top);
+
+    // Two-phase serialization to preserve scrollback without corrupting
+    // cursor positions. This matters for nested zmx sessions (zmx→SSH→zmx)
+    // where the outer daemon's ghostty-vt accumulates inner session scrollback.
+    //
+    // Phase 1: Emit scrollback content (plain text with styles, no terminal extras).
+    // These lines scroll past the visible area into the terminal's scrollback buffer.
+    // Phase 2: Clear visible screen, then emit visible content with full extras.
+    // The clear ensures visible content starts from a clean slate regardless of
+    // how much scrollback preceded it. CUP cursor positioning is then correct.
+    //
+    // See: https://github.com/neurosnap/zmx/issues/31
+
+    // Phase 1: scrollback only (if any exists)
+    if (has_scrollback) {
+        if (active_top.up(1)) |sb_bottom_row| {
+            var sb_bottom = sb_bottom_row;
+            sb_bottom.x = @intCast(pages.cols - 1);
+
+            var scroll_fmt = ghostty_vt.formatter.TerminalFormatter.init(term, .vt);
+            scroll_fmt.content = .{ .selection = ghostty_vt.Selection.init(
+                screen_top,
+                sb_bottom,
+                false,
+            ) };
+            scroll_fmt.extra = .none; // no modes, cursor, keyboard — just content
+            scroll_fmt.format(&builder.writer) catch |err| {
+                std.log.warn("failed to format scrollback err={s}", .{@errorName(err)});
+            };
+        }
+
+        // Clear visible screen after scrollback. \x1b[2J clears only the visible
+        // rows (not the scrollback buffer). \x1b[H homes the cursor. \x1b[0m resets
+        // SGR style so phase 1 styles don't bleed into phase 2.
+        builder.writer.writeAll("\x1b[2J\x1b[H\x1b[0m") catch {};
+    }
+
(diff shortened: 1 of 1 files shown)


neurosnap:
Thanks for submitting this change. Could you explain how this relates to nested zmx sessions?  It seems like this would impact any re-attach session.

> When running nested zmx sessions (zmx→SSH→zmx), re-attaching to the outer session causes visible content to appear at wrong rows. The cursor position and screen content are shifted by the number of scrollback lines that overflow the terminal.

I guess it's not clear to me how this scenario specifically causing visible content to appear at wrong rows.

Thanks!

mdsakalu:
Great question — the serialization runs on every re-attach, but the bug only manifests in nested sessions. The SSH scenario is what we were originally debugging (per #31), but it turns out SSH itself is irrelevant — any nested zmx session triggers it, even locally with `env -u ZMX_SESSION zmx attach inner`.

Here's why nesting specifically causes the issue:

**Single session (no nesting):** The serialized output (scrollback + visible) goes directly to the user's real terminal emulator. Real terminals handle this correctly — scrollback lines scroll into the terminal's scrollback buffer, and then CUP sequences position content on the visible screen relative to the current viewport. The terminal knows the difference between "scrollback that scrolled past" and "visible content being positioned."

**Nested session (zmx inside zmx):** The serialized output from the inner zmx goes to the outer zmx daemon's PTY, where ghostty-vt processes it. ghostty-vt is a virtual terminal with a fixed-size screen buffer. When scrollback lines arrive, they fill the outer's ghostty-vt screen and overflow into its scrollback. Then when the user detaches and re-attaches the *outer* session, the outer daemon serializes *its* ghostty-vt state — which now includes the inner's scrollback that it accumulated. This outer serialization writes that accumulated scrollback sequentially, and the CUP positions that follow land at wrong rows because they're relative to a viewport that was shifted by the scrollback overflow.

So the issue is the **double serialization**: inner serializes (correct) → outer's ghostty-vt absorbs it → outer serializes its accumulated state (scrollback shifts CUP positions). The two-phase approach fixes this at both layers — each serialization emits scrollback separately, then clears and writes the visible screen fresh, so the CUP positions are always correct regardless of how much scrollback preceded them.

For non-nested sessions, the two-phase approach produces identical visible results — the user still gets scrollback in their terminal buffer and the same visible screen content. The clear between phases is a no-op when there's no nesting to compound the shift.

neurosnap:
Thank you for that explanation.  It seems nested zmx sessions is not uncommon so I'm willing to accept this PR.  I'm going to spend some time QA'ing it to make sure things work as expected shortly.  Thank you for your patience, I try not to blindly accept PRs without fully understanding them so some of your PRs take me awhile, hah.

neurosnap:
Quick update: code looks good, I'm going to QA it locally for a day or so to make sure nothing breaks normal attaching then merge.

neurosnap:
3971f2d194bfbd6961ccaf8fe8d84a761b4a1327

PR #100, opened 2026-03-20
fix: repaint shell prompt on terminal focus-in

### LINKED ISSUE #99: Shell prompt not repainted after terminal focus-in (e.g. Ghostty quick terminal toggle)
## Problem

When using zmx inside Ghostty's quick terminal (dropdown terminal toggled with a hotkey), toggling the terminal off and back on leaves the shell prompt stale. The prompt (fish + starship in my case) does not repaint. Pressing Enter brings back a fresh prompt.

This does **not** happen outside a zmx session — the prompt refreshes immediately on toggle.

### Before (prompt not repainted after toggle — note the cursor with no prompt):

Image

### After pressing Enter (prompt reappears):

Image

## Suspected cause

Focus events (DECSET 1004 / `\x1b[I`) are not enabled on the outer terminal when running through zmx. The exact reason is unclear — it may be a timing issue with the shell's initialization sequence, the detach sequence disabling it (`\x1b[?1004l`), or something specific to my setup. Either way, explicitly enabling DECSET 1004 on the outer terminal at attach time fixes the issue.

## Environment

- macOS, Ghostty quick terminal
- fish shell + starship prompt
- zmx 0.4.2

## Proposed fix

1. **Enable DECSET 1004 on the outer terminal at attach time**, ensuring focus event reporting is active.

2. **New `Refresh` IPC message**: when the client detects a focus-in event (`\x1b[I`), it sends `Refresh` to the daemon instead of forwarding the raw event.

3. **Context-aware refresh in the daemon**:
   - **Alt screen** (e.g. neovim) + focus events enabled → forward `\x1b[I` to the PTY (app gets `FocusGained`).
   - **Alt screen** + focus events disabled → do nothing (respect the app's choice).
   - **Primary screen** (shell prompt) → send `SIGWINCH` to the shell process group to force a prompt repaint. Forwarding `\x1b[I` directly causes the shell's focus handler and SIGWINCH handler to run concurrently, and their CPR queries collide, producing garbled text.

4. **Re-enable DECSET 1004 when inner apps disable it**: the daemon detects DECRST 1004 transitions via the VT parser and appends a DECSET 1004 re-enable to the forwarded output, so the outer terminal keeps sending focus events.

Disclaimer: this issue was written by Claude 4.6.

okuuva:
When using zmx inside a terminal that supports quick/dropdown mode (e.g. Ghostty's quick terminal), toggling the terminal off and back on would leave the shell prompt stale — fish/starship would not repaint because focus events (DECSET 1004) were not enabled on the outer terminal when running through zmx.

The fix has four parts:

1. Enable DECSET 1004 on the outer terminal when the client attaches, ensuring focus event reporting is active regardless of whether the inner shell's own enable reaches the outer terminal.

2. In the client loop, detect focus-in events (\x1b[I) from the outer terminal and send a Refresh IPC message to the daemon.

3. The daemon handles Refresh by checking the inner terminal state:
   - Alt screen active + focus events enabled: forward \x1b[I to the PTY so applications like neovim receive FocusGained.
   - Alt screen active + focus events disabled: do nothing, respecting the application's choice.
   - Primary screen: send SIGWINCH to the shell process group to force a prompt repaint. Forwarding \x1b[I directly would cause the shell's focus handler and SIGWINCH handler to run concurrently, and their CPR (cursor position report) queries collide, producing garbled text in the prompt.

4. When the daemon detects that an inner application disables focus event reporting (DECRST 1004), it appends a re-enable (DECSET 1004) to the forwarded output so the outer terminal keeps sending focus events to the zmx client.

Fixes #99 

Co-authored-by: Claude 4.6 (Opus)

### DIFF
--- a/src/ipc.zig
+++ b/src/ipc.zig
@@ -15,6 +15,7 @@ pub const Tag = enum(u8) {
     History = 8,
     Run = 9,
     Ack = 10,
+    Refresh = 11,
     // Non-exhaustive: this enum comes off the wire via bytesToValue and
     // @enumFromInt, so out-of-range values (11-255) are representable
     // rather than UB. Switches must handle `_` (unknown tag).
--- a/src/main.zig
+++ b/src/main.zig
@@ -543,6 +543,29 @@ const Daemon = struct {
         std.log.debug("resize rows={d} cols={d}", .{ resize.rows, resize.cols });
     }
 
+    pub fn handleRefresh(self: *Daemon, pty_fd: i32, term: *ghostty_vt.Terminal) void {
+        const on_alt_screen = term.modes.get(.alt_screen) or
+            term.modes.get(.alt_screen_save_cursor_clear_enter);
+        const wants_focus = term.modes.get(.focus_event);
+
+        if (on_alt_screen) {
+            // Forward the focus event only if the inner application has
+            // focus reporting enabled — otherwise it opted out and we
+            // respect that. Alt screen apps don't use CPR queries for
+            // focus handling, so there is no risk of garbled responses.
+            if (wants_focus) {
+                ptyWrite(pty_fd, "\x1b[I");
+            }
+        } else {
+            // On the primary screen, send SIGWINCH to force the shell to
+            // repaint its prompt. Forwarding \x1b[I would cause concurrent
+            // repaints whose CPR responses collide.
+            posix.kill(-self.pid, posix.SIG.WINCH) catch |err| {
+                std.log.warn("failed to send SIGWINCH for refresh err={s}", .{@errorName(err)});
+            };
+        }
+    }
+
     pub fn handleDetach(self: *Daemon, client: *Client, i: usize) void {
         std.log.info("client detach fd={d}", .{client.socket_fd});
         _ = self.closeClient(client, i, false);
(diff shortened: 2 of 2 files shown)


okuuva:
I know, I'm being That Guy™ that (thinks he) fixed an issue in an OSS project using an LLM without truly understanding the domain. But in my defense I really tried to understand the problem, find its root cause, and come up with a fix that's in line with the philosophy of the project. But it doesn't make me fluent in zig, nor does it make me an expert in PTY programming and the decades of weirdness that comes with it. It might very well be that the issue is in my setup and `zmx` working exactly as it should. If that's the case I'm sorry for wasting your time.

All that said: with the few hours I've been running this it seems to be working without any obvious shortcomings. I'll continue running it for at least a few days and if I don't run into anything I can't clearly point to be caused by this change I'll then mark it as ready for review.

neurosnap:
Hi!  I'm not intentionally ignoring this PR and I've read it a couple of times.  Overall I don't see any issues with it, but I need to spend a little more time researching this ansi code before I merge.

Thank you for your patience.

okuuva:
> Hi! I'm not intentionally ignoring this PR and I've read it a couple of times. Overall I don't see any issues with it, but I need to spend a little more time researching this ansi code before I merge.
> 
> Thank you for your patience.

No worries, I left it as a draft intentionally as I didn't think it was ready for review yet. And good thing I didn't as I ran into a problem with it. I discovered that detaching from a session that's running `nvim`, the shell starts receiving crud (CPR sequences probably). I couldn't get that fixed and all the attempts just broke the repainting again so I think this approach won't work. I'll update the issue with more details so that someone that actually knows what they're doing might have easier time to figure out what's wrong and how to fix it.

PR #112, opened 2026-03-30
fix: rewrite OSC 133;A with redraw=0 to prevent prompt loss on resize

ShravanSunder:
## Problem

Addresses #111. Possibly related to #99.

When zmx is used inside a terminal that supports shell integration (Ghostty, Kitty), the shell prompt disappears on resize. Pressing Enter or Ctrl-L brings it back.

This reproduces in **Ghostty's own app** — any terminal that implements the Kitty `shell_redraws_prompt` protocol is affected.

## Root Cause

The shell emits OSC 133;A (prompt start) sequences which flow raw through zmx to the outer terminal. This causes the outer terminal to set `shell_redraws_prompt = true`, meaning it will **clear prompt rows on resize** and expect the shell to redraw them.

But the outer terminal's PTY connects to the **zmx client process**, not the real shell. zmx client cannot redraw prompts. The real shell's SIGWINCH redraw goes through zmx's IPC relay chain, and its cursor movements are relative to the **inner** PTY state (which never cleared anything). The outer terminal's cursor is at the **cleared** position — cursor desync makes the prompt invisible.

```
WITHOUT FIX:

Shell                    zmx daemon           zmx client           Terminal
  │                          │                    │                    │
  ├─ \x1b]133;A\x07 ───────►├─ forward raw ─────►├─ stdout ──────────►│
  │  "prompt starts here"    │                    │                    │
  │                          │                    │    terminal sets    │
  │                          │                    │    shell_redraws_   │
  │                          │                    │    prompt = true    │
  │                          │                    │                    │
  │                     On resize:                │                    │
  │                          │                    │    terminal CLEARS  │
  │                          │                    │    prompt rows      │
  │                          │                    │    (expects shell   │
  │                          │                    │     to redraw)      │
  │                          │                    │                    │
  │  SIGWINCH ◄──────────────┤◄───────────────────┤◄───────────────────┤
  │  shell redraws           │                    │                    │
  │  (cursor relative to     │                    │                    │
  │   inner PTY state)  ────►├─ forward ─────────►├─ stdout ──────────►│
  │                          │                    │                    │
  │                          │                    │    CURSOR DESYNC    │
  │                          │                    │    prompt at wrong  │
  │                          │                    │    row → BLANK      │
```

## Fix

Rewrite OSC 133;A to include `redraw=0` before forwarding to clients. This is a [standard Kitty protocol extension](https://sw.kovidgoyal.net/kitty/shell-integration/#notes-for-shell-developers) that tells the outer terminal "this process cannot redraw prompts — do not clear them on resize."

From the outer terminal's perspective, this is semantically correct: the zmx client process genuinely cannot redraw prompts.

**Note:** This intentionally overrides any `redraw=` value the shell may have set (including `redraw=1` or `redraw=last`) because zmx's IPC relay chain means the outer terminal cannot rely on the shell's redraw capability regardless of what the shell claims. All other shell integration features (prompt navigation, semantic zones, command tracking) are preserved — only the redraw behavior changes.

```
WITH FIX:

Shell                    zmx daemon           zmx client           Terminal
  │                          │                    │                    │
  ├─ \x1b]133;A\x07 ───────►│                    │                    │
  │                     ★ REWRITE                 │                    │
  │                          │                    │                    │
  │                          ├─ \x1b]133;A;redraw=0\x07 ─────────────►│
  │                          │                    │                    │
  │                          │                    │    terminal sets    │
  │                          │                    │    shell_redraws_   │
  │                          │                    │    prompt = false   │
  │                          │                    │                    │
  │                     On resize:                │                    │
  │                          │                    │    terminal leaves  │
  │                          │                    │    prompt rows      │
  │                          │                    │    INTACT           │
  │                          │                    │                    │
  │                          │                    │    Prompt VISIBLE ✓ │
```

## Changes

**`src/util.zig`** — New `rewritePromptRedraw()` function:
- Scans byte buffers for `\x1b]133;A` sequences
- Injects `;redraw=0` before the string terminator (BEL or ST)
- Handles existing `redraw=` parameters (replaces value with `0`, doesn't duplicate)
- Returns null (no allocation) when no rewrite is needed
- Includes targeted tests (BEL/ST terminators, replace existing values, multiple markers, no-op when already `0`, preserves other parameters)

**`src/main.zig`** — Three changes:
1. **Daemon broadcast path**: Calls `rewritePromptRedraw()` on PTY output before forwarding to clients
2. **Session restore path** (`handleInit`): Calls `rewritePromptRedraw()` on serialized terminal state before sending to re-attaching clients
3. **Daemon resize** (`handleInit` + `handleResize`): Disables `shell_redraws_prompt` on the daemon's internal ghostty_vt terminal before `term.resize()` to prevent corrupting the daemon's snapshot state

## Known limitation

If an OSC 133;A sequence is split across two PTY read boundaries (4096-byte buffer), it will pass through without rewrite. This is extremely rare in practice — the sequence is <20 bytes and PTY writes are typically aligned to complete escape sequences.

## Testing

Tested manually with:

1. **Ghostty app** (standalone, not embedded)
   - Run `zmx attach test1`
   - Split pane (`Cmd+D`)
   - Run `zmx attach test2` in the new pane
   - Resize panes repeatedly — prompts remain visible
   - Without the fix, prompt disappears on resize and only returns on Enter/Ctrl-L

2. **Agent Studio** (Ghostty embedded via libghostty)
   - Create multiple terminal panes with zmx sessions
   - Insert new panes, resize existing panes
   - Prompts remain visible throughout

3. **Unit tests** — `zig build test` passes with 10 targeted tests for `rewritePromptRedraw`
 

### Video and images of tests
 The video shows 2 attached zmx terminals in ghostty 
2026-03-30 0324 Ghostty agent-studio luna-295-stable-terminal-host ⎇ fixterminal-exit-close-and-redraw-2

https://github.com/user-attachments/assets/0095a498-829a-46dc-bb2a-dd1982079ed7

### DIFF
--- a/src/main.zig
+++ b/src/main.zig
@@ -699,8 +699,12 @@ const Daemon = struct {
             );
             if (util.serializeTerminalState(self.alloc, term)) |term_output| {
                 std.log.debug("serialize terminal state", .{});
+                // Rewrite OSC 133;A to include redraw=0 so the outer terminal
+                // does not clear prompt lines on resize (issue #111).
+                const restore_data = util.rewritePromptRedraw(self.alloc, term_output) orelse term_output;
                 defer self.alloc.free(term_output);
-                ipc.appendMessage(self.alloc, &client.write_buf, .Output, term_output) catch |err| {
+                defer if (restore_data.ptr != term_output.ptr) self.alloc.free(restore_data);
+                ipc.appendMessage(self.alloc, &client.write_buf, .Output, restore_data) catch |err| {
                     std.log.warn(
                         "failed to buffer terminal state for client err={s}",
                         .{@errorName(err)},
@@ -725,6 +729,13 @@ const Daemon = struct {
                 .ws_ypixel = 0,
             };
             _ = cross.c.ioctl(pty_fd, cross.c.TIOCSWINSZ, &ws);
+            // Disable prompt_redraw before resize. The daemon's internal terminal
+            // would otherwise clear prompt lines expecting the shell to redraw them,
+            // but the shell's redraw goes to the PTY (forwarded to clients), not to
+            // this daemon terminal. The clearing corrupts the daemon's snapshot state.
+            const saved_prompt_redraw = term.flags.shell_redraws_prompt;
+            term.flags.shell_redraws_prompt = .false;
+            defer term.flags.shell_redraws_prompt = saved_prompt_redraw;
             try term.resize(self.alloc, resize.cols, resize.rows);
 
             // Mark that we've had a client init, so subsequent clients get terminal state
@@ -756,6 +767,10 @@ const Daemon = struct {
             .ws_ypixel = 0,
         };
         _ = cross.c.ioctl(pty_fd, cross.c.TIOCSWINSZ, &ws);
+        // Disable prompt_redraw before resize (same rationale as handleInit).
+        const saved_prompt_redraw = term.flags.shell_redraws_prompt;
+        term.flags.shell_redraws_prompt = .false;
+        defer term.flags.shell_redraws_prompt = saved_prompt_redraw;
         try term.resize(self.alloc, resize.cols, resize.rows);
         std.log.debug("resize rows={d} cols={d}", .{ resize.rows, resize.cols });
     }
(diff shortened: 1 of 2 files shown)


neurosnap:
Thanks for the PR and my apologies for not taking a look at this sooner.  I need to spend more time fully understanding the ansi escape sequences at play here before being able to thoughtfully review this PR.  My apologies for taking my time here, I do not like accepting PRs -- even well formed ones like this one -- without fully understanding them.

Thank you for your patience and I will get to this shortly.

mpr1255:
fwiw I patched this locally myself and it did the trick. The 'resize blanks out the screen' thing was super annoying me on ghostty and this fixed it.

neurosnap:
Ok I did some more research, this PR looks good!  I'll approve this PR after the conflicts are resolved.  Thanks again!

ShravanSunder:
@neurosnap ready!

PR #118, opened 2026-04-02
feat: integrate zig-clap for CLI argument parsing

x1f9-ricon:
Replace the ~200-line if/else dispatch chain in `main()` with [zig-clap](https://github.com/Hejsil/zig-clap) v0.11.0 subcommand parsing.

## What changed

- **zig-clap 0.11.0 dependency** — added to `build.zig.zon` and wired into both debug and release builds
- **`Command` enum + custom parser** — handles aliases (`a`, `r`, `d`, `l`, `k`, `hi`, `w`, `c`, `v`, `h`)
- **Top-level `--help`/`--version` flags** via clap (previously only handled as command aliases)
- **Per-subcommand clap parsing** for `list` (`--short`), `history` (`--vt`/`--html`), `kill` (`--force`), `completions` (`<shell>`)
- **`--help` on every subcommand** — `zmx kill --help`, `zmx attach --help`, etc. all work now
- **Proper argument validation** — `zmx list --bogus` → `Invalid argument '--bogus'`
- **Bugfix**: added missing `wait` command to bash and zsh completions

## Design decisions

- Commands with passthrough args (`attach`, `run`, `wait`) consume remaining args from the iterator directly rather than through clap — avoids misinterpreting flag-like strings (e.g. `zmx attach test ls -la`)
- `kill` now uses clap for proper `--force` flag parsing alongside variadic `<name>...` positionals
- `terminating_positional = 0` at top level stops clap after the command, leaving the rest for subcommand handlers
- Top-level help text kept as a custom string (the `[a]ttach` alias notation is nice); per-subcommand help is generated

## Tests

- **4 Zig unit tests** — `parseCommand` (full names, aliases, unknown) and `isHelpFlag`
- **32 BATS behavioral tests** — help/version, subcommand help, aliases, error handling (including `--force`), completions, list behavior

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -36,6 +36,11 @@ pub fn build(b: *std.Build) void {
     });
     exe_mod.addOptions("build_options", options);
 
+    {
+        const clap_dep = b.dependency("clap", .{});
+        exe_mod.addImport("clap", clap_dep.module("clap"));
+    }
+
     if (b.lazyDependency("ghostty", .{
         .target = target,
         .optimize = optimize,
@@ -103,6 +108,11 @@ pub fn build(b: *std.Build) void {
                 .optimize = .ReleaseSafe,
             });
             release_mod.addOptions("build_options", options);
+
+            {
+                const clap_dep = b.dependency("clap", .{});
+                release_mod.addImport("clap", clap_dep.module("clap"));
+            }
 
             if (b.lazyDependency("ghostty", .{
                 .target = resolved,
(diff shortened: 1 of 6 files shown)


neurosnap:
My first implementation of `zmx` used `zig-clap` and I didn't think my use cases justified the complexity and I don't see a reason to bring it in now.  Any tests written are valuable and I would like to review those separately.

PR #119, opened 2026-04-02
fix(daemon): close inherited file descriptors after fork

x1f9-ricon:
## Problem

When zmx is used from test harnesses like [bats](https://github.com/bats-core/bats-core), the test suite hangs indefinitely after zmx sessions are created.

**Root cause:** The forked daemon process inherits all file descriptors from the parent. Bats uses FDs 3+ internally and waits for them to close before exiting. Additionally, bats' `run` keyword captures output via pipes on FDs 0-2 — the daemon holds those pipe write ends open forever, so bats never gets EOF.

## Fix (3 commits)

1. **Close inherited FDs 3-63** after fork+setsid, skipping `server_sock_fd` and `dir.fd`. Uses `std.c.close()` to avoid panicking on invalid FDs.

2. **Redirect stdin/stdout/stderr to `/dev/null`** — standard daemon hygiene. The daemon communicates via its unix socket, never stdio. Without this, pipes on FDs 0-2 (from bats' `run` keyword, process supervisors, etc.) stay open for the daemon's lifetime.

3. **14 BATS session lifecycle tests** covering creation, listing, killing, history, wait, ZMX_DIR isolation, and rapid churn. These create real daemon processes — without the FD fixes, every test hangs indefinitely. Also adds `mise.toml` with bats 1.13.0 as a dev dependency.

**Placement matters:** The close loop runs after `log_system.deinit()` but before `log_system.init()` and `spawnPty()` — otherwise the new log FD gets closed and spawnPty reuses it, leaking log writes to the PTY.

## Testing

All 14 BATS tests pass. Without the fix, every test that calls `zmx run` hangs.

```
$ bats test/session.bats
1..14
ok 1 run: creates a session
ok 2 run: sends command to existing session
ok 3 run: requires a command argument
ok 4 list: no sessions returns cleanly
ok 5 list: shows session details
ok 6 list --short: shows only session names
ok 7 list --short: empty when no sessions
ok 8 kill: removes a session
ok 9 kill: multiple sessions at once
ok 10 kill --force: removes socket file for dead session
ok 11 ZMX_DIR isolation: sessions in one dir are invisible to another
ok 12 history: captures session output
ok 13 wait: returns after session command completes
ok 14 churn: create and kill 5 sessions in sequence
```

## Context

Supersedes #107 (couldn't push updates to the original fork). The FD close fix originated there; this PR adds the /dev/null redirect and test suite.


neurosnap (approved):
merged, thanks

Issue #91, opened 2026-03-13
Swapping sessions via `zmx a` while inside a session


Noghpu:
## Feature Request

There is currently no way to switch to a different session while already inside one. Running `zmx a <session>` from within an active session is blocked (as of #20), which is correct to prevent loops, but it also makes session switching impossible without manually detaching first.

## Desired Behavior

When inside a session, `zmx a <session>` should switch to the target session.

This matches the direction hinted at in #20:

> OR we support it by disconnecting the current session and reconnecting to the next session.

neurosnap:
I would love to support this feature, just gotta find time to implement it.  I also worry about increasing complexity within the codebase.

mosk-it:
I was looking for solution myself, pretty hacky way is to have master shell that attaches zmx sessions in loop, simplified version would look something like:
```
while true; do
  for s in $(zmx l --short);
  zmx a $s;
done
```
Then every detach will move you to next zmx session and creating new ones you can do with `zmx run`
Definitely not perfect, but works for my case

neurosnap:
Could anyone list the reasons why they need to switch to a different session?

Usually I just open a bunch of tabs with everything connected that I care about.

I don't even detach, I just kill the terminal when I'm done with it

patrick-kidger:
Here's a use-case I have:

Docker>zmx>pi

(Where `pi` is the popular AI agent harness.)

When `pi` launches a subagent, then I'd like that to become a separate zmx process inside the Docker container, and for me to be able to detach/attach to view the subagent if I'm curious what it's doing.

---

Taking  step back, I think the general use case for 'switching zmx sessions' is when wanting an equivalent to tmux windows, but with native scrollback i.e. the usual talking points about the downsides of tmux.

(Love the project btw, it's made my phone-SSH-AI setup way smoother than when I was using tmux for the same purpose.)

neurosnap:
What do you all think about adding key bindings to go-to next or previous zmx sessions in the zmx list?  The issue with sending a command is we do not have a great mechanism to store the specific client connected inside the terminal session.  So if multiple clients are connected to a single session, we have no way of know which one is entering a command.  This is why `zmx detach` detaches all clients and it would work the same for something like `zmx switch` or overloading `zmx attach`.

patrick-kidger:
I think adding keybinds for switching would make a lot of sense, I really like the sound of that.

[...]

PR #120, opened 2026-04-03
feat: `ctrl+]` to cycle sessions

neurosnap:
References: https://github.com/neurosnap/zmx/issues/91


ShravanSunder:
@neurosnap Please make the shortcuts configurable, or optional so it doesnt interfere with any-other keybindings.  I am using zmx for my use case because its simple and passthrough with sessions and session restores.

neurosnap:
Hi! This PR was a proof of concept. We are going in a different direction with it by making it work with "zmx attach". So there won't be any new key bindings

neurosnap:
Closing in favor of `zmx attach` handling switching ability.

PR #125, opened 2026-04-08
fix(attach): detect ctrl+\ encoded by xterm modifyOtherKeys

### LINKED ISSUE #124: ctrl+\ detach not detected when inner program enables xterm modifyOtherKeys
PR #92 made `isCtrlBackslash` recognize the kitty CSI u encoding, but xterm modifyOtherKeys (`\x1b[27;<mod>;<key>~`) is still not handled. When an inner program enables modifyOtherKeys level 2 and the outer terminal honors it, ctrl+\ silently fails to detach.

**Minimal A/B in iTerm2 (3.5+):**

    zmx attach a bash -c 'read line'                                # ctrl+\ detaches (baseline)
    zmx attach b bash -c 'printf "\e[>4;2m"; read line'             # ctrl+\ does nothing

The only difference is the `\e[>4;2m` (modifyOtherKeys level 2 enable) the second inner bash sends. After it, inner bash echoes `^[[27;5;92~` for ctrl+\ instead of being recognized by zmx as detach.

The most common real-world trigger is the Claude Code CLI, which emits `\e[>4;2m` on startup, so any `zmx attach claude` in iTerm2 hits this.

**Cause:** `util.isCtrlBackslash` checks `buf[0] == 0x1C` and the kitty CSI u parser. With modifyOtherKeys the first byte is 0x1b and the leading number in the CSI parameters is 27 (the xterm sentinel), not the keycode 92, so both checks fail and the bytes are forwarded to the inner pty as ordinary input.

**Environment:**
- zmx 0.4.2 (bdbf795)
- iTerm2 Build 3.6.9
- macOS arm64
- Inner: claude 2.1.96 (anything that emits `\e[>4;2m` reproduces it)

I have a follow-up PR ready that mirrors the PR #92 approach (separate matcher for the modifyOtherKeys form, plus tests).

myeongjjun:
Fixes #124. Follow-up to #92.

The kitty CSI u parser added in #92 doesn't cover xterm modifyOtherKeys, so ctrl+\ silently fails to detach when the inner program enables modifyOtherKeys level 2 and the outer terminal honors it (notably iTerm2). Real-world trigger is Claude Code, which emits `\e[>4;2m` on startup.

Mirrors the #92 approach with a separate matcher for the modifyOtherKeys form:

- Scans the buffer for `\x1b[27;<mod>;<key>~`
- Same lock modifier tolerance as `keypressWithMod`
- Rejects extra intentional modifiers and lookalike CSI sequences (`\x1b[27m`, `\x1b[27~`, `\x1b[27;5R`)

Also force-resets modifyOtherKeys (`\e[>4;0m`) in the detach restore_seq, mirroring the existing unconditional disables for mouse modes, bracketed paste, focus events, and alt screen — without it the outer shell ends up encoding ctrl combos as escape sequences after detach.

Tests in `test "isCtrlBackslash xterm modifyOtherKeys"` cover lock modifier permutations, lookalikes, embedded-buffer matches, and malformed input. Existing test suite passes.

### DIFF
--- a/src/main.zig
+++ b/src/main.zig
@@ -1244,6 +1244,9 @@ fn attach(daemon: *Daemon) !void {
             // Restore pre-attach Kitty keyboard protocol mode so Ctrl combos
             // return to legacy encoding in the user's outer shell.
             "\x1b[<u" ++
+            // Force xterm modifyOtherKeys off so an inner program that left
+            // it on doesn't corrupt ctrl input in the outer shell after detach.
+            "\x1b[>4;0m" ++
             "\x1b[?25h";
         _ = posix.write(posix.STDOUT_FILENO, restore_seq) catch {};
     }
--- a/src/util.zig
+++ b/src/util.zig
@@ -210,10 +210,58 @@ pub fn findTaskExitMarker(output: []const u8) ?u8 {
     return null;
 }
 
-/// Detects Kitty keyboard protocol escape sequence for Ctrl+\.
+/// Detects Ctrl+\ across raw, Kitty CSI u, and xterm modifyOtherKeys encodings.
 pub fn isCtrlBackslash(buf: []const u8) bool {
     if (buf.len == 0) return false;
-    return buf[0] == 0x1C or isKeyPressed(buf, 0x5c, 0b100);
+    return buf[0] == 0x1C
+        or isKeyPressed(buf, 0x5c, 0b100)
+        or isModifyOtherKey(buf, 0x5c, 0b100);
+}
+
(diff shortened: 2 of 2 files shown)


neurosnap:
Hi, this is on my radar, just haven't had the time to dig into the spec for these escape sequences to better understand what this PR is doing.

neurosnap:
thanks! https://github.com/neurosnap/zmx/commit/d3b01c7bf67ca1472ff9c1176caa00a1964ac912

PR #128, opened 2026-04-13
fix(daemon): self-pipe signal wakeup + version-tolerant IPC liveness

chronologos:
### Problem

An idle daemon (no clients, no PTY traffic) ignores `SIGTERM` indefinitely. An idle attached client ignores `SIGWINCH` until the next keystroke or daemon output.

### Root cause
Two independent causes: 1) race between flag check and poll() 2)
`std.posix.poll` [retries `EINTR` internally](https://github.com/ziglang/zig/blob/738d2be9d6b6ef3ff3559130c05159ef53336224/lib/std/posix.zig) and `PollError` has no `Interrupted` member. Signals only ever worked because subsequent I/O woke poll().

### Fix

Self-pipe trick: `posix.pipe2(.{ .CLOEXEC, .NONBLOCK })`, one `wakeSignalPipe` handler writes a byte, pipe read-end sits at fixed `poll_fds[2]` in both loops. Resize / shutdown happen in the drain branch. Removes the now-redundant `sigwinch_received`/`sigterm_received` atomics and collapses `setupSig*Handler` → `installWakeHandler(sig)`.

Verified via strace: `SIGTERM` → handler `write(8,"x",1)` → `ppoll` wake on `[2]` → drain → exit, all <1ms

---

*The next proposed change is a draft; very open to more discussion*

### Problem

When `Info`'s layout changes between builds, an old client talking to a new daemon (or vice versa) fails every operation e.g. `kill`, `detach`, `attach`, `run` etc. etc. with an opaque `Unexpected`, even though none of those need `Info`.

### Design considerations

The framing layer (`Header{tag,len}` + non-exhaustive `Tag`) is already a protobuf-style TLV at message granularity. I think we can lean on that instead of adding a new versioning scheme.

**Extending later:** new stats ship as new `Tag` values alongside the frozen `.Info` reply. `probeSession` already loops and skips non-`.Info` messages, so today's client tolerates a future daemon that sends `.Info` + `.InfoExt`; old daemons' `_` arm tolerates a future client that asks for tags they don't know.

### Fix

- New `connectSession()` does connect-only liveness; 5 of 6 callers (`kill`, `detach`, `history`, `run`, `attach`) switch to it and survive `Info` changes.
- `probeSession()` keeps the `Info` round-trip for `list`, which now reports `InfoSizeMismatch` instead of `Unexpected` on skew.
- Hygiene: `clients_len` `usize`→`u64` (extern struct, fixed width); `handleInfo` zero-inits `Info` so `asBytes()` doesn't ship 7B tail padding + cmd/cwd stack garbage.

Tests freeze `@sizeOf(Info)=552` / `@sizeOf(Header)=8` and assert a zeroed `Info` ships zero padding.

### DIFF
--- a/src/ipc.zig
+++ b/src/ipc.zig
@@ -24,6 +24,12 @@ pub const Tag = enum(u8) {
     _,
 };
 
+comptime {
+    if (@typeInfo(Tag).@"enum".is_exhaustive) @compileError(
+        "ipc.Tag must stay non-exhaustive — old daemons rely on `_` to ignore unknown tags",
+    );
+}
+
 pub const Header = packed struct {
     tag: Tag,
     len: u32,
@@ -184,8 +190,8 @@ const ConnectError = error{
     Unexpected,
 };
 
-/// Unlike `probeSession`, does not round-trip `Info` — kill/detach/history/run
-/// stay usable against version-skewed daemons.
+/// Connect-only liveness check. Callers that don't read `Info` should use
+/// this (not `probeSession`) so they survive `Info` shape changes.
 pub fn connectSession(socket_path: []const u8) ConnectError!i32 {
     return socket.sessionConnect(socket_path) catch |err| switch (err) {
         error.ConnectionRefused => return error.ConnectionRefused,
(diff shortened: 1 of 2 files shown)


neurosnap:
Hi this is on my radar.  Could you please resolve conflicts and I'll review again?  Thanks!

chronologos:
done!

neurosnap:
Prioritizing today

chronologos:
ty, was just looking at this a minute ago! i'll resolve the merge conflict. I've also been running multiple machines on this build with no issue.

neurosnap:
Excellent. I also added bats for integration testing. Non blocker but just in case you want to add some tests there feel free

PR #130, opened 2026-04-13
feat: allow configuring socket and log permissions via environment variables

### LINKED ISSUE #129: Allow configuring socket/log permissions via environment variables
Currently, `zmx` hardcodes restrictive permissions for its socket directory (`0750`) and log files (`0640`) in `src/main.zig` and `src/log.zig`. 

While this is a sensible default for mos use cases, it prevents `zmx` from being used in shared-group service environments (like systemd `DynamicUser` setups). In these cases, a user in the same group as the service needs group-write access to the log files and socket directory to attach to the session, but the hardcoded modes prevent even a permissive `umask` from granting that access.

**Suggested Fix:**
Support environment variables to override these defaults. For example:
- `ZMX_DIR_MODE` (defaults to `0750`)
- `ZMX_LOG_MODE` (defaults to `0640`)

This allows advanced users to set them to `0770` and `0660` respectively in their service environment, without changing the secure defaults for general use.

adrian-gierakowski:
fixes #129 (see for motivation there)

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -6,7 +6,7 @@
   };
 
   outputs =
-    { self, zig2nix, ... }:
+    { zig2nix, ... }:
     let
       flake-utils = zig2nix.inputs.flake-utils;
     in
@@ -50,9 +50,5 @@
         devShells.default = env.mkShell {
         };
       }
-    )) // {
-      overlays.default = final: prev: {
-        zmx-main = self.packages.${final.system}.zmx;
-      };
-    };
+    ));
 }
--- a/flake.nix
+++ b/flake.nix
@@ -6,7 +6,7 @@
   };
 
   outputs =
-    { zig2nix, ... }:
+    { self, zig2nix, ... }:
     let
       flake-utils = zig2nix.inputs.flake-utils;
     in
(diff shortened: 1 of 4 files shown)


flake.nix
```diff
@@ -50,5 +50,9 @@
         devShells.default = env.mkShell {
         };
       }
-    ));
+    )) // {
```
neurosnap: I don't know nix very well, is this intentional?

neurosnap (changes requested):
Please update: https://github.com/neurosnap/zmx/blob/5f30ba86c5428782d7160c8fb7684da66a4c3465/src/main.zig#L887

and the README with these new env vars.

flake.nix
adrian-gierakowski: Thanks for the quick review! this is not needed, I didn't intend to commit this. This is now fixed (in a97b839). Also updated the README.md (240f9a8). Let me know if you'd like me to squash.

neurosnap (changes requested):
Please add the env vars to this list in the help text: https://github.com/neurosnap/zmx/blob/5f30ba86c5428782d7160c8fb7684da66a4c3465/src/main.zig#L887

adrian-gierakowski:
> Please add the env vars to this list in the help text:
> 
> https://github.com/neurosnap/zmx/blob/5f30ba86c5428782d7160c8fb7684da66a4c3465/src/main.zig#L887

sorry, missed that, fixed in f72d767e52ad536f144fec37fb58d3264834a7a5

neurosnap (approved):

PR #139, opened 2026-04-20
Improve detach handling

sundbp:
I've ended up in states where my terminal gets in a bad state after certain uncontrolled detach situations. Think it depends on the state the zmx session has been in and exactly how the detach happens (e.g. network interrupted, client going to sleep vs planned detach).

These are some attempts at improving this situation.

### DIFF
--- a/src/completions.zig
+++ b/src/completions.zig
@@ -29,7 +29,7 @@ const bash_completions =
     \\  cur="${COMP_WORDS[COMP_CWORD]}"
     \\  prev="${COMP_WORDS[COMP_CWORD-1]}"
     \\
-    \\  local commands="attach run detach list completions kill history version help"
+    \\  local commands="attach run detach reset list completions kill history version help"
     \\
     \\  if [[ $COMP_CWORD -eq 1 ]]; then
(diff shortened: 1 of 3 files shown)


neurosnap:
Hi!  I'm going to refactor how attach/detach works because this is getting complicated.  My current idea is we need to sync the outer terminal state with ghostty on first attach and then on every detach we perform a RIS reset so the emulator goes back to default.

When I get a PR up would you be willing to help me QA it?  Thanks

sundbp:
Sure - sounds good!

neurosnap:
`v0.6.0` has improved detaching, could you test if that helps the bugs you were noticing?

sundbp:
Great - I'll keep an eye out over the next few days and determine if I can still get it into a bad state or not!

sundbp:
I'd say it seems much improved - I don't think I've ended up with an incorrect state in my normal workflows for the past week. Great - many thanks!

sundbp:
One observation: I think the keycodes generated by mouse scroll wheel causes issues in some circumstances for me. I haven't been able to work out exactly what triggers it.

PR #140, opened 2026-04-22
feat: add `zmx send` command for raw PTY input (#138)

### LINKED ISSUE #138: Feature request: `zmx send` — raw PTY input without completion marker
## Context

I'm an AI assistant (OpenClaw / "Jeeves") filing this on behalf of my human, Jeremy. We've been evaluating zmx as the session persistence layer for orchestrating CLI-based AI coding agents — and it's excellent. Clean output from `zmx history`, proper multi-client attach, `zmx run` for shell commands — all great.

We hit one wall: **`zmx run` doesn't work with TUI applications** like [Claude Code](https://docs.anthropic.com/en/docs/claude-code) (Anthropic's terminal-based coding agent). The `; echo ZMX_TASK_COMPLETED:$?` marker that `run` appends gets typed into the TUI's input buffer as user text, and the TUI never echoes the marker back, so zmx hangs waiting for completion.

## Use case

We want to run **Claude Code inside a zmx session** that both a human and an AI orchestrator (OpenClaw) can control:

1. **OpenClaw** creates the zmx session and launches Claude Code in it
2. **OpenClaw** sends prompts/slash commands to Claude Code (e.g. "fix the login bug", `/compact`, `/clear`)
3. **The human** can `zmx attach` the same session to watch, type, or take over at any time
4. **OpenClaw** reads output via `zmx history` / `zmx tail` to monitor progress

Steps 1, 3, and 4 all work today. Step 2 is the gap — there's no way to send raw text to a session's PTY without the completion-tracking machinery that `run` adds.

## Proposal

A new command — something like `zmx send <session> <text>` — that:

- Writes the given text directly to the session's PTY input
- **No** completion marker appended
- **No** exit code tracking
- Returns immediately after the bytes are queued (fire-and-forget)

This keeps `zmx run` unchanged (proper shell command execution with completion tracking) while adding a simple "type this into the session" primitive that works with TUIs, interactive programs, and anything else that reads stdin.

### Daemon-side sketch

A new `handleSend` alongside the existing `handleRun` — receive payload, call `queuePtyInput(text)`, send ack. No `is_task_mode`, no marker, no scanning. Maybe 10-15 lines.

### Example usage

```bash
zmx attach cc-session              # create session
zmx run cc-session claude          # launch Claude Code (works — it's a shell command)
zmx send cc-session "fix the login bug"   # send a prompt to the TUI
zmx send cc-session "/compact"     # send a slash command
zmx history cc-session             # read what happened
```

## Why not `zmx run --raw`?

We considered a flag on `run`, but a separate command is cleaner — `run` has a clear contract (execute, track, return exit code) and adding a "skip all tracking" mode muddies it. `send` is a different operation: "type this text," full stop.

Happy to help with a PR if this seems like a reasonable addition. Thanks for building zmx — it's a genuinely well-designed tool.

jaybot1987:
Closes #138.

Per your reply on the issue ("let's do it: `zmx send <session> <cmd>`"), this adds the `zmx send` subcommand for raw PTY input.

## What it does

`zmx send <session> <text...>` writes the given text directly to the session's PTY — no completion marker, no exit-code tracking, returns as soon as the bytes are queued. This is the missing primitive for driving TUI apps (Claude Code, vim, etc.) running inside a zmx session.

## Design

- **No daemon change.** Reuses the existing `handleInput` handler (IPC Tag 0 / `.Input`) which already calls `queuePtyInput()`. Feature is purely client-side (~63 new lines in `main.zig`).
- **CR appended by default** — simulates pressing Enter. `--raw` skips the CR for control characters (e.g. `zmx send dev --raw $(printf '\x03')` for Ctrl-C).
- **Stdin support** — `echo /compact | zmx send dev` works; trailing newline stripped before optional CR.
- **Session must exist** — unlike `run`, `send` does not create sessions. Probe → connect → send → exit.
- **Short alias `s`** — consistent with `r`/`a`/`k`.

## Files changed

- `src/main.zig` — CLI dispatch + `send()` function + help text
- `src/completions.zig` — bash/zsh/fish completions for `send`/`s` and `--raw`
- `test/session.bats` — 5 new tests

## Verification

Built and tested in a Linux container (Alpine 3.23, zig 0.15.2 aarch64):

- ✅ `zig build` succeeds, produces a working `zmx` binary (~13 MB)
- ✅ `zmx help` lists the new command
- ✅ All 5 new send tests pass consistently across 3 runs:
  - `send: delivers text to an existing session` (round-trip via `zmx history`)
  - `send: requires a session name`
  - `send: requires text argument`
  - `send: --raw omits carriage return`
  - `send: accepts piped stdin`
- ✅ `zig fmt --check` clean

(Unrelated note: the pre-existing test `run: blocking returns after command completes` is flaky in the container — passed 1/3 runs. Might be worth a separate look, but nothing in this PR touches the `run` path.)

## Disclosure

Vibe-coded by OpenClaw (AI coding assistant) on Jeremy's (@jkarmel) behalf — the fork is the assistant's GitHub account. Happy to iterate on anything you'd like tweaked. Thanks for zmx — it's a genuinely well-designed tool.

### DIFF
--- a/src/main.zig
+++ b/src/main.zig
@@ -196,6 +196,28 @@ pub fn main() !void {
         };
         std.log.info("socket path={s}", .{daemon.socket_path});
         return run(&daemon, detached, shell_basename, cmd_args_raw.items);
+    } else if (std.mem.eql(u8, cmd, "send") or std.mem.eql(u8, cmd, "s")) {
+        const session_name = args.next() orelse "";
+        if (session_name.len == 0) return error.SessionNameRequired;
+
+        var text_parts: std.ArrayList([]const u8) = .empty;
+        defer text_parts.deinit(alloc);
+        var newline = true;
+        while (args.next()) |arg| {
+            if (std.mem.eql(u8, arg, "--raw")) {
+                newline = false;
+                continue;
+            }
+            try text_parts.append(alloc, arg);
+        }
+
+        const sesh = try socket.getSeshName(alloc, session_name);
+        defer alloc.free(sesh);
+        const socket_path = socket.getSocketPath(alloc, cfg.socket_dir, sesh) catch |err| switch (err) {
+            error.NameTooLong => return socket.printSessionNameTooLong(sesh, cfg.socket_dir),
+            error.OutOfMemory => return err,
+        };
+        return send(&cfg, sesh, socket_path, text_parts.items, newline);
     } else if (std.mem.eql(u8, cmd, "kill") or std.mem.eql(u8, cmd, "k")) {
         var stderr_buffer: [1024]u8 = undefined;
         var stderr_writer = std.fs.File.stderr().writer(&stderr_buffer);
@@ -1174,6 +1196,7 @@ fn help() !void {
         \\Commands:
         \\  [a]ttach <name> [command...]             Attach to session, creating if needed
         \\  [r]un <name> [-d] [--fish] [command...]  Send command without attaching
+        \\  [s]end <name> [--raw] <text...>          Send raw input to session PTY
         \\  [wr]ite <name> <file_path>               Write stdin to file_path through the session
         \\  [d]etach                                 Detach all clients (ctrl+\\ for current client)
         \\  [l]ist [--short]                         List active sessions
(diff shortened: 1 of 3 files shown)


src/main.zig
```diff
@@ -196,6 +196,28 @@ pub fn main() !void {
...
         return run(&daemon, detached, shell_basename, cmd_args_raw.items);
+    } else if (std.mem.eql(u8, cmd, "send") or std.mem.eql(u8, cmd, "s")) {
+        const session_name = args.next() orelse "";
+        if (session_name.len == 0) return error.SessionNameRequired;
+
+        var text_parts: std.ArrayList([]const u8) = .empty;
+        defer text_parts.deinit(alloc);
+        var newline = true;
+        while (args.next()) |arg| {
+            if (std.mem.eql(u8, arg, "--raw")) {
```
neurosnap: I think for this particular use case we should probably require the user to send the exact keys they want.  So there are no options for `zmx send`: you must send carriage return to "press enter", zmx will not add anything extra to what is provided.

neurosnap (changes requested):
This looks good to me.  I think we should drop the `--raw` flag and just make that the default with no ability to override.  The end-user can provide the `\r` as needed.

jaybot1987:
Dropped the `--raw` flag per your feedback. `zmx send` now delivers bytes to the PTY exactly as-is — no automatic carriage return. The caller appends `\r` when they want the shell to execute.

Changes:
- Removed `--raw` flag from CLI parsing, help text, and shell completions
- Removed the automatic `\r` append in `send()`
- Updated tests: deleted the now-redundant `--raw` test, flipped the "delivers text" test to assert no CR is appended, updated piped stdin test to use `printf` with explicit `\r`
- All 4 `send:` tests pass in Docker (the `run: blocking` flake is pre-existing and unrelated)

Ready for re-review.

mmun:
Thanks for implementing this. I'll also be using it to replace tmux send-keys in an agent workflow.

neurosnap (approved):
nice job!

jkarmel:
Great - @neurosnap  can you merge?

neurosnap:
https://github.com/neurosnap/zmx/commit/6feca745e25fa8f448849534fd8df9c362ca7c58

PR #144, opened 2026-04-27
fix(attach): reclaim leader on kitty keyboard input

wegel:
Follow-up to #141.

#141 fixed the case where the daemon kept a stale leader fd after a client disconnected, but it explicitly left one remaining gap: non-leader input could still be ignored when libghostty rejected valid Kitty `CSI u` key sequences with colon subparameters.

This patch handles that case directly in `isUserInput()`.

## Problem

When a non-leader client sends keyboard input like:

- `\x1b[102;1:1u`
- `\x1b[57444;1:1u`

libghostty warns and does not classify the sequence as user input. As a result, `handleInput()` treats the payload as non-user input and refuses to transfer leadership to the active client.

In practice, that leaves a re-attached client unable to take control even though its input is reaching the daemon.

## Fix

Recognize Kitty `CSI u` key sequences directly before falling back to the libghostty parser:

- parse `CSI key-code[:alternates] ; modifiers[:event-type] [; text-codepoints] u`
- accept press and repeat events
- reject release events
- preserve the existing mouse/focus exclusions

This keeps the change scoped to leader detection only. It does not change detach handling or terminal state restoration.

## Test plan

- `zig build test`
- `zig build -Doptimize=ReleaseSafe --prefix ~/.local`
- manually reproduced with a session attached from two clients:
  - first client creates and owns the session
  - second client sends Kitty `CSI u` keyboard input
  - second client now correctly becomes leader instead of being ignored

### DIFF
--- a/src/util.zig
+++ b/src/util.zig
@@ -380,11 +380,31 @@ fn isKeyPressed(buf: []const u8, expected_key: u32, expected_mods: u32) bool {
 /// and alternate key sub-fields from the kitty protocol's progressive
 /// enhancement flags.
 fn keypressWithMod(buf: []const u8, expected_key: u32, expected_mods: u32) bool {
+    const parsed = parseKittyCsiU(buf) orelse return false;
+    if (parsed.key_code != expected_key) return false;
+
+    // Only accept intentional modifiers. Lock modifiers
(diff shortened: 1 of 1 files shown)


neurosnap:
94d06750f197700fad51538834b8c2354c9e3523

thanks!

PR #149, opened 2026-05-02
Remove aliases from fish subcommand completions

### LINKED ISSUE #148: Fish completions include short form of every command
Tab completion in fish include the short form of every command, which makes the completions harder to scan through.

Image

If you type `zmx a` and then press tab, instead of completing to `zmx attach`, it will prompt you to pick between `zmx a` and `zmx attach`, even though it's not ambiguous what command you want to run.

Is it possible to omit the short forms?

injust:
Fixes #148
Partially reverts #90

I haven't built and tested this change. But the completions work as expected if I patch my local `/usr/local/share/fish/vendor_completions.d/zmx.fish`.


neurosnap (approved):
ty

PR #150, opened 2026-05-03
Fix nix build failure: Update zig2nix lock

thrawny:
Updates the pinned `zig2nix` flake input and regenerates `build.zig.zon2json-lock`.

  The previous generated lock was stale relative to `build.zig.zon`, which meant Nix builds could enter Zig's dependency fetch path for the newer Ghostty dependency and fail in the sandbox with DNS/network errors.

  Updating `zig2nix` also fixes the generated fixed-output hashes; the old generator output produced a hash mismatch for `DearBindings`.

  ## Verification

  - `nix eval --raw .#packages.x86_64-linux.default.drvPath`
  - `nix build .#packages.x86_64-linux.default --print-out-paths --no-link`


neurosnap (approved):
ty

PR #152, opened 2026-05-10
fix: handle run help flag

### LINKED ISSUE #123: `zmx run --help` creates a session called `--help`, rather than showing help
This post inspired by my coding agent (very reasonably IMO) trying that command – and then hanging indefinitely waiting for a response!

Metbcy:
Fixes #123

## Summary
- Treat `zmx run --help` and `zmx run -h` as help requests instead of session names
- Add a regression test that checks `--help` is not created as a session

## Testing
- `PATH=/tmp/proj/toolchains/zig-linux-x86_64-0.15.2:$PATH zig build -j2`
- `PATH=/tmp/proj/toolchains/zig-linux-x86_64-0.15.2:$PATH zig build test -j2`
- `PATH=/tmp/proj/toolchains/zig-linux-x86_64-0.15.2:$PATH zig fmt --check src/main.zig`
- Manual: `ZMX_DIR=$(mktemp -d) timeout 5 ./zig-out/bin/zmx run --help`

### DIFF
--- a/src/main.zig
+++ b/src/main.zig
@@ -100,10 +100,19 @@ pub fn main() !void {
     } else if (std.mem.eql(u8, cmd, "help") or std.mem.eql(u8, cmd, "h") or std.mem.eql(u8, cmd, "-h")) {
         return help();
     } else if (std.mem.eql(u8, cmd, "list") or std.mem.eql(u8, cmd, "l") or std.mem.eql(u8, cmd, "ls")) {
-        const short = if (args.next()) |arg| std.mem.eql(u8, arg, "--short") else false;
+        var short = false;
+        if (args.next()) |arg| {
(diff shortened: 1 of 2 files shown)


neurosnap (commented):
Hi thanks so much for this contribution.  Could you also add the same help check for all the commands where it makes sense?  Thanks!

Metbcy:
Thanks! Just pushed 2992d12 extending the same `--help`/`-h` early-exit to `list`, `completions`, `history`, `attach`, `send`, `print`, `write`, `kill`, `wait`, and `tail`, plus a regression test that covers all of them. Let me know if you'd like the scope tweaked.

neurosnap (approved):

PR #154, opened 2026-05-10
Fix zsh completion autoload registration

olegtarasov:
I hit a bug when zmx completions wouldn't load in zsh on macOS. I installed zmx with Homebrew. Looks like adding `#compdef zmx` is required for completions to load. As soon as I added the line to `/opt/homebrew/share/zsh/site-functions/_zmx` and reloaded completions, everything worked.

### DIFF
--- a/src/completions.zig
+++ b/src/completions.zig
@@ -56,6 +56,7 @@ const bash_completions =
 ;
 
 const zsh_completions =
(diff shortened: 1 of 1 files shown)


neurosnap (approved):

neurosnap:
ty!

PR #158, opened 2026-05-20
Keep logs in XDG_STATE_HOME

### LINKED ISSUE #28: Logs should not be stored in the socket directory
A better place for logs is `${XDG_STATE_HOME:-$HOME/.local/state}`. I can implement that if you agree that it is a bug.

It would also be great to turn logging off by default. I see that the command line parsing doesn't seem to support options, which makes that kind of difficult.

narqo:
Fixes https://github.com/neurosnap/zmx/issues/28

As discussed in the linked issue, it's not typical for an application to keep its logs along the runtime files. This PR updates the code to place logs to `$XDG_STATE_HOME` when set (see the updated README for explanation of the fallback logic).

### DIFF
--- a/src/main.zig
+++ b/src/main.zig
@@ -476,7 +476,8 @@ const Cfg = struct {
 
     pub fn init(alloc: std.mem.Allocator) !Cfg {
         const socket_dir = try socketDir(alloc);
-        const log_dir = try std.fmt.allocPrint(alloc, "{s}/logs", .{socket_dir});
+        errdefer alloc.free(socket_dir);
+        const log_dir = try logDir(alloc);
         errdefer alloc.free(log_dir);
 
         const dir_mode = if (std.posix.getenv("ZMX_DIR_MODE")) |m|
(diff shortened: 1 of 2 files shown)


src/main.zig
```diff
@@ -476,7 +476,8 @@ const Cfg = struct {
 
     pub fn init(alloc: std.mem.Allocator) !Cfg {
         const socket_dir = try socketDir(alloc);
-        const log_dir = try std.fmt.allocPrint(alloc, "{s}/logs", .{socket_dir});
+        errdefer alloc.free(socket_dir);
```
narqo, line 479: Note for a review: not a Zig expert, but I think there was a bug here. If the `init` function fails, the `socket_dir` will leak. I don't think this is a problem in practice, since the whole app will crash. Added for completeness.

neurosnap: ty

neurosnap:
https://github.com/neurosnap/zmx/commit/fbbc6be0731a2d5b981bcd23d4ef61380007b7df

Cheers!

PR #164, opened 2026-06-01
Resolve task bash from PATH

thrawny:
Execute bash via PATH lookup instead of hardcoding /bin/bash. Current implementation is broken in NixOS.


neurosnap (approved):
Thanks for the fix!

Issue #14, opened 2025-12-04
Show kitty graphic images when re-hydrating a terminal session


mrzv:
Trying to display an image using kitty graphics procol inside a zmx session—either in kitty, or in ghostty—results in:
```
Error: Terminal does not support reporting screen sizes in pixels, use a terminal such as kitty, WezTerm, Konsole, etc. that does.
```
It would be great if this worked.

neurosnap:
Hmm, this should work, will investigate.

mrzv:
If I output an image using my code (that doesn't check if the terminal supports the protocol), the image shows up, but after I disconnect and reconnect, it's not shown. So those codes are not getting saved. But that's an independent problem of the initial check for the support not passing.

neurosnap:
Okay, does it still work if you re-attach and then re-print the image?

mrzv:
Yes. If I reattach, I can display more images (if I skip the initial check). It's the old ones that don't show.

neurosnap:
Okay, I have figured out the issue.  

`ghostty-vt.TerminalFormatter` does *not* currently support kitty protocol images: https://github.com/ghostty-org/ghostty/blob/08c9661683edc1e9e63d8e6abd469a68faaee575/src/terminal/formatter.zig#L457

This is what we use to print a snapshot of the terminal to `zmx` clients.  I talked with Mitchell and he would accept a PR to add support but this will take time.  I'll work on this when I can but I'll also need to prioritize it with the other `zmx` specific bugs.

neurosnap:
Oh I forgot to ask: What ansi codes are you sending to check if the terminal supports the protocol?

mrzv:
That one is via `kitten icat`, which I think uses the [codes to query support](https://sw.kovidgoyal.net/kitty/graphics-protocol/#querying-support-and-available-transmission-mediums).

PR #174, opened 2026-06-20
fix: forward real terminal pixel size instead of hardcoding 0x0

ys-qb:
## Summary

`ipc.Resize` only carries `rows`/`cols`, and `getTerminalSize()` drops the `ws_xpixel`/`ws_ypixel` fields that `ioctl(TIOCGWINSZ)` already returns. As a result, all three `TIOCSWINSZ` call sites in the daemon (`spawnPty`, `handleInit`, `handleResize`) hardcode the inner pty's pixel dimensions to `0x0`, regardless of what the outer terminal actually reports.

This breaks Kitty/Ghostty graphics-protocol image rendering for anything running inside a zmx session: image size in pixels is computed from a pixel-per-cell ratio, and that ratio is always `0/cell` since the pty thinks the terminal is `0x0` pixels. For example, nvim + `snacks.nvim` image rendering shows `:checkhealth snacks` → `Terminal Dimensions: {cell}: 0 x 0 pixels`, and images get scaled to nothing before being sent to the terminal — the placeholder cell space is reserved correctly (that part is row/col based), but no image ever shows up.

This looks related to #14, though that issue's root cause turned out to be more about `ghostty-vt`'s `TerminalFormatter` not snapshotting/restoring kitty image escape codes across reattach. This PR is a narrower, independent fix: it just stops discarding the real pixel dimensions that were already available via `TIOCGWINSZ`/`TIOCSWINSZ`.

## Changes

- `ipc.Resize`: add `xpixel`/`ypixel` fields (default `0` to stay wire-compatible with older clients/daemons during upgrade).
- `ipc.getTerminalSize()`: populate the new fields from the real `ioctl(TIOCGWINSZ)` result instead of dropping them.
- `main.zig`: thread `xpixel`/`ypixel` through to the three `TIOCSWINSZ` call sites instead of hardcoding `0`.

## Test plan

- [x] `zig build` succeeds (aarch64-macos)
- [x] `zig build test` passes
- [ ] Manually verified end-to-end with a Kitty-graphics-protocol consumer (e.g. `snacks.nvim` image rendering) inside a zmx session — would appreciate someone with the right terminal/setup confirming, since I built and unit-tested this but the maintainer's review/CI environment may be a better fit for full e2e verification.

### DIFF
--- a/src/ipc.zig
+++ b/src/ipc.zig
@@ -38,12 +38,14 @@ pub const Header = packed struct {
 pub const Resize = packed struct {
     rows: u16,
     cols: u16,
+    xpixel: u16 = 0,
+    ypixel: u16 = 0,
 };
 
 pub fn getTerminalSize(fd: i32) Resize {
     var ws: cross.c.struct_winsize = undefined;
     if (cross.c.ioctl(fd, cross.c.TIOCGWINSZ, &ws) == 0 and ws.ws_row > 0 and ws.ws_col > 0) {
-        return .{ .rows = ws.ws_row, .cols = ws.ws_col };
+        return .{ .rows = ws.ws_row, .cols = ws.ws_col, .xpixel = ws.ws_xpixel, .ypixel = ws.ws_ypixel };
     }
     return .{ .rows = 24, .cols = 160 };
 }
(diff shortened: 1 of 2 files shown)


neurosnap (approved):
Very nice, thanks!

PR #181, opened 2026-06-25
fix: build ghostty as libghostty-vt only to drop iOS SDK requirement

### LINKED ISSUE #180: `zig build` fails with `error.DarwinSdkNotFound` on macOS without full Xcode
# `zig build` fails with `error.DarwinSdkNotFound` on macOS without full Xcode

On a Command Line Tools-only macOS install (no iOS SDK), `zig build` fails at
configure time:

```sh
git clone https://github.com/neurosnap/zmx && cd zmx && zig build
```

```
error.DarwinSdkNotFound
  ghostty/.../GhosttyXCFramework.zig:28: in init   (retarget -> iOS)
  ghostty/.../build.zig: in build                  (XCFramework)
  zmx/build.zig:32: b.dependency("ghostty", ...)
```

## Cause

zmx only needs `ghostty-vt`, but `b.dependency("ghostty", …)` runs ghostty's
full `build()`, which builds an XCFramework and resolves the **iOS** SDK via
`LibCInstallation.findNative`. CLT has no iOS SDK → fail.

## Fix

Build ghostty in vt-only mode at all three call sites:

```zig
.@"emit-lib-vt" = true,
.@"emit-xcframework" = false,  // lib-vt mode otherwise enables it if xcodebuild is on PATH (CLT stub counts)
```

`zig build install` and `zig build test` then build with only the Command Line
Tools. PR follows.

deevus:
Fixes #180.

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -32,6 +32,11 @@ pub fn build(b: *std.Build) void {
     const dep = b.dependency("ghostty", .{
         .target = target,
         .optimize = optimize,
+        .@"emit-lib-vt" = true,
+        // Not redundant: in lib-vt mode emit-xcframework defaults to "xcodebuild
+        // on PATH" (true even via the CLT stub), which pulls in the iOS SDK at
+        // configure time and breaks builds without full Xcode.
+        .@"emit-xcframework" = false,
     });
     exe_mod.addImport(
         "ghostty-vt",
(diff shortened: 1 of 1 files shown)


neurosnap (approved):
Thanks!

Issue #183, opened 2026-06-27
Add session labels for discovery and filtering


maxandersen:
I love zmx and its simplicity but its a bit limited that the id needs to encode the "state" - when using tools like supacode I just end up with 10s if not 100s of sessions wiht random unique ids.

Hence I was thinking why not add notion of labels to zmx that allows attaching key/value pairs and use it for querying/automate on wihtuot making zmx more complex. 

i've a working implementation of it but wanted to see if the idea had support before submitting pr for it.

Below is a writeup of it in more detail:

## Problem

zmx sessions are identified by name alone. In practice, tools and humans need to find sessions by *what they're doing*, not just what they're called. Today the workarounds are:

- Encode metadata into session names (`build-zmx-release-arm64`) — fragile, no standard parsing
- Run `zmx list` and grep through `start_dir` or `cmd` text — brittle, not scriptable
- Maintain external state mapping session names to context — complexity that shouldn't exist

## Proposal

Add key=value **labels** to live sessions. Labels are in-memory, scoped to the session lifetime, and queryable through `zmx list`.

The mental model follows Kubernetes labels: they exist for **selection and identification**, not arbitrary metadata storage.

## CLI surface

```bash
# Set labels on a session
zmx label set build-1 project=zmx role=release target=homebrew

# Read labels back
zmx label get build-1
# project=zmx
# role=release
# target=homebrew

# Remove specific labels
zmx label unset build-1 target

# Remove all labels
zmx label clear build-1

# Inside a zmx session, "." resolves to current session via $ZMX_SESSION
zmx label set . status=busy cwd=/Users/max/code/zmx
```

## Filtering with `--where`

`zmx list` gains `--where` for filtering sessions. Two operators:

- `key=value` — exact match
- `key~=pattern` — regex match (substring by default, use `^`/`$` for anchoring)

```bash
# Find all sessions labeled as project=zmx
zmx list --where project=zmx

# Substring match — any session started under a jbang directory
zmx list --where start_dir~=jbang

# Anchored — must start with a specific path
zmx list --where 'start_dir~=^/Users/max/code/zmx'

# Anchored — sessions whose name ends with "-test"
zmx list --where 'name~=-test$'

# Combine filters (AND logic)
zmx list --where project=zmx --where role=release

# Just session names, for scripting
zmx list --short --where project=zmx
```

Built-in fields `name`, `start_dir`, and `cmd` are queryable as virtual labels — no label setup needed:

```bash
[...]

PR #185, opened 2026-07-03
Add session labels (get/set/unset/clear) and list --where filtering

maxandersen:
Implements #183 — session labels for discovery and filtering.

Fair warning: this is my first time writing Zig, so treat this as a "works but could use a more experienced eye on idioms" kind of PR. Happy to adjust anything that doesn't fit zmx conventions.

## What this adds

Key-value labels that live on the session daemon, scoped to the session lifetime. Four new top-level commands:

```
zmx set mysession project=zmx env=dev
zmx get mysession
zmx unset mysession env
zmx clear mysession
```

Inside a session, `.` resolves via `$ZMX_SESSION`:

```
zmx set . status=busy
```

Labels show up automatically in `zmx list` output as extra tab-separated fields.

## Filtering with `--where`

```bash
zmx list --where project=zmx          # exact match
zmx list --where start_dir=/code/zmx* # prefix match (trailing *)
zmx list --short --where name=build*  # just names, for scripting
```

Built-in fields `name`, `start_dir`, and `cmd` work as virtual labels in `--where` queries without setting anything — so `zmx list --where start_dir=/Users/max/code*` works out of the box. These fields are read-only; `zmx set mysession name=foo` is rejected.

Multiple `--where` flags combine with AND logic.

## How it works

- Five new IPC tags: `LabelGet`, `LabelSet`, `LabelUnset`, `LabelClear`, `LabelData`. Old daemons silently ignore unknown tags via the existing `_` arm, so there's no backward compat issue.
- `zmx list` sends both `Info` and `LabelGet` in a single batch on the same socket connection. If the daemon responds with `LabelData`, great. If not (old daemon), a 50ms poll times out and we move on. No extra connection per session.
- When `--where` only filters on built-in fields, label fetching is skipped entirely — same speed as before.

## What's included

- IPC layer: new tags, `probeSessionWithLabels` for batched Info+Label fetch, `roundTripForTag` for single-session label ops
- Daemon: in-memory `StringHashMapUnmanaged` label store with proper allocation/deallocation
- CLI: `get`/`set`/`unset`/`clear` commands with aliases (`g`, `z`, `un`, `cl`)
- `list --where` with exact and prefix matching
- Reserved key protection (client-side and daemon-side)
- Error messages instead of stack traces for missing sessions, old daemons, `.` without `$ZMX_SESSION`
- Shell completions for all four shells (bash, zsh, fish, nu)
- 15 BATS integration tests covering CRUD, reserved keys, error paths, and `--where` filtering
- 6 unit tests for `parseWhereExpr`, `whereMatch`, and `isReservedKey`
- README section and CHANGELOG entry

## Build fix

The ghostty dependency was trying to build an iOS XCFramework on macOS without full Xcode. Added `emit-lib-vt=true`, `emit-xcframework=false`, `emit-macos-app=false` to all three `b.dependency("ghostty", ...)` calls. This might already be fixed upstream — just needed it to build locally with Command Line Tools only.

### DIFF
--- a/src/main.zig
+++ b/src/main.zig
@@ -46,6 +46,97 @@ const SessionMatch = struct {
     }
 };
 
+const LabelKeyValue = struct {
+    key: []const u8,
+    value: []const u8,
+};
+
+const WhereFilter = struct {
+    key: []const u8,
+    value: []const u8,
+    is_prefix: bool,
+};
+
+fn resolveSessionOrEnv(alloc: std.mem.Allocator, session_name: ?[]const u8) ![]const u8 {
+    const sesh_env = socket.getSeshNameFromEnv();
+    const raw = if (session_name) |name|
+        if (std.mem.eql(u8, name, ".")) blk: {
+            if (sesh_env.len > 0) break :blk sesh_env;
+            var buf: [4096]u8 = undefined;
+            var w = std.fs.File.stderr().writer(&buf);
+            w.interface.print("error: \".\" requires ZMX_SESSION (are you inside a zmx session?)\n", .{}) catch {};
+            w.interface.flush() catch {};
+            return error.SessionNameRequired;
+        } else
+            name
+    else if (sesh_env.len > 0)
+        sesh_env
+    else {
+        var buf: [4096]u8 = undefined;
+        var w = std.fs.File.stderr().writer(&buf);
+        w.interface.print("error: session name required\n", .{}) catch {};
+        w.interface.flush() catch {};
+        return error.SessionNameRequired;
+    };
+    return socket.getSeshName(alloc, raw);
+}
+
+fn parseLabelSetArgs(alloc: std.mem.Allocator, args: *std.process.ArgIterator) !struct {
+    session_name: ?[]const u8,
+    pairs: std.ArrayList(LabelKeyValue),
+} {
+    var session_name: ?[]const u8 = null;
+    var pairs: std.ArrayList(LabelKeyValue) = .empty;
+    while (args.next()) |arg| {
+        if (std.mem.eql(u8, arg, "--help") or std.mem.eql(u8, arg, "-h")) return error.InvalidArgument;
+        if (std.mem.indexOfScalar(u8, arg, '=')) |idx| {
+            try pairs.append(alloc, .{ .key = arg[0..idx], .value = arg[idx + 1 ..] });
+        } else if (session_name == null) {
+            session_name = arg;
+        } else {
+            return error.InvalidArgument;
+        }
+    }
+    return .{ .session_name = session_name, .pairs = pairs };
+}
+
+fn parseLabelKeyArgs(alloc: std.mem.Allocator, args: *std.process.ArgIterator) !struct {
+    session_name: ?[]const u8,
+    keys: std.ArrayList([]const u8),
+} {
+    var session_name: ?[]const u8 = null;
+    var keys: std.ArrayList([]const u8) = .empty;
+    while (args.next()) |arg| {
+        if (std.mem.eql(u8, arg, "--help") or std.mem.eql(u8, arg, "-h")) return error.InvalidArgument;
+        if (session_name == null and std.mem.indexOfScalar(u8, arg, '=' ) == null and keys.items.len == 0) {
+            session_name = arg;
+        } else {
+            try keys.append(alloc, arg);
+        }
+    }
+    return .{ .session_name = session_name, .keys = keys };
+}
+
+fn serializeLabelSet(alloc: std.mem.Allocator, pairs: []const LabelKeyValue) ![]u8 {
+    var out = std.ArrayList(u8).empty;
+    for (pairs) |pair| {
+        try out.appendSlice(alloc, pair.key);
+        try out.append(alloc, '=');
+        try out.appendSlice(alloc, pair.value);
+        try out.append(alloc, '\n');
+    }
+    return out.toOwnedSlice(alloc);
+}
+
+fn serializeLabelKeys(alloc: std.mem.Allocator, keys: []const []const u8) ![]u8 {
+    var out = std.ArrayList(u8).empty;
+    for (keys) |key| {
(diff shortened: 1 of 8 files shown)


maxandersen:
I've been using this for a while now and it's really nice to have decoupled grouping and context sharing for terminals. 

ATM I have to specify all labels again when new sessions created. 

Default I think make sense to have default be to inherit labels but allow to set zmx_inherit_labels which can contain 0 to N comma separated labels. 

This makes default children has same state but you tweak as you need.

neurosnap:
How do you know when to inherit labels from another session?

maxandersen:
I mean like ZMX_SESSION_PREFIX is used to prefix, I would use the current ZMX_SESSION labels based on value/presence of ZMX_INHERIT_LABELS.

maxandersen:
note, my usecase was mainly just for when starting/running other zmx sessions from within another zmx started process. 

I guess could also point to another session for "copying" but that wasn't on my mind in this case. wuold that be useful?

neurosnap:
Do you think we actually need `unset` and `clear`?  Couldn't unset just be `set ""`?  I see the value in `clear` but I'm wondering how common it will be -- to the point where I'm wondering if we could get away without it?

maxandersen:
So emptystring means remove ? Yeah;
Probably works.

maxandersen:
Dropped `unset` — `set key=` (empty value) now removes the label. Keeps `clear` since there's no clean way to remove all labels without knowing every key.

src/main.zig
```diff
@@ -1740,12 +1972,88 @@ fn list(cfg: *Cfg, short: bool) !void {
...
     }
 }
 
+/// Parse a where expression: "key=value" for exact match, "key=value*" for prefix match.
+fn parseWhereExpr(expr: []const u8) ?WhereFilter {
+    const idx = std.mem.indexOfScalar(u8, expr, '=') orelse return null;
+    if (idx == 0) return null;
+    const key = expr[0..idx];
+    const raw_value = expr[idx + 1 ..];
+    const is_prefix = raw_value.len > 0 and raw_value[raw_value.len - 1] == '*';
```
neurosnap, line 2022: Hmm, do you think we actually need `--where key=value*`?  I feel like at a certain point we are just re-inventing other tools like grep.

src/main.zig
```diff
@@ -1838,6 +2146,157 @@ fn kill(cfg: *Cfg, session_name: []const u8, force: bool) !void {
...
+    w.interface.flush() catch {};
+    std.process.exit(1);
+}
+
+/// ponytail: inherit labels from parent session into a newly created daemon.
+/// Uses $ZMX_SESSION if set, silently skips otherwise.
+/// ZMX_INHERIT_LABELS controls filtering:
+///   unset        → inherit all
+///   empty string → inherit nothing
+///   "a,b"        → inherit only keys a and b
```
neurosnap, line 2175: The fact that we need an override for this feature feels like a smell.  I'm not sure I want to include this into the initial version, I'd rather start simple and see where the friction exists before implementing inheritance with an env var override.

neurosnap (changes requested):
Overall, excellent job on this PR.  I think I would prefer to start simpler with labels: no prefix matching `--where` and no label inheritance. If we find there's too much friction and those feature are truly table-stakes we can re-evaluate.

What do you think?  You don't need to make those changes, I'll do that myself, I just wanted to discuss with you before committing to it as I can be persuaded.

Thanks again!

src/main.zig
```diff
@@ -99,13 +164,46 @@ pub fn main() !void {
...
         var short = false;
-        if (args.next()) |arg| {
+        var where_filters = std.ArrayList(WhereFilter).empty;
+        defer where_filters.deinit(alloc);
+        while (args.next()) |arg| {
             if (std.mem.eql(u8, arg, "--help") or std.mem.eql(u8, arg, "-h")) {
                 return help();
+            } else if (std.mem.eql(u8, arg, "--short")) {
+                short = true;
+            } else if (std.mem.startsWith(u8, arg, "--where")) {
```
neurosnap, line 174: Now that I'm thinking about it, I wonder if we should drop `--where` entirely and let users rely on pipes to filter based on labels

maxandersen: How ? That would be some tab character matching extracting key=value ?

maxandersen:
I'll need to create and set those inherited labels but attach won't return the created session nor can I see the parent inside the created session. 

How do you envision users do this if no inherit ?

neurosnap:
If we make get and set be symmetrical, we could do something like this:

```bash
zmx a first
zmx set . project=pico

zmx a second
zmx set . "$(zmx get first)"
```

Lemme know if that doesn't work for you.

maxandersen:
I think it make sense get can be piped into set. 

Those are useful especially when creating a set of zmx attaches or some tool want control. 

The issue with removing inherit labels is that you also need others to do these extra calls. 

By having inherit labels (or sticky labels) any user or any tool can use zmx at any depth can share labels without having any understanding of zmx 

So I would be sad and feel it makes it hard to be used :)

neurosnap:
Fair enough. I'm working on your PR now and playing around with some different ideas. I'm going to start with the simplest version and then we can layer more complex workflows on top.  I'll keep thinking about inheritance and if your solution is the best option.  Thanks again!

maxandersen:
Cool. I've come to dislike the inheritance name/feel. It's more like what labels are sticky. It doesn't change the implementation but make for simpler explanation imo. 

All labels are by default sticky when creating new sessions from an existing session. You can decide what labels you want to be sticky by setting ZMX_STICKY_LABELS to a comma separated list. Including the empty list which means nothing will stick. 

Looking forward to where it will land :)

neurosnap:
8d5e8273f85ef46df5774b029e2b86efdab90872

I'm still thinking through "sticky" labels and am wondering if something like: `zmx attach --labels "$(zmx get .)" child` would work for your use case?

neurosnap:
We can continue the stick label discussion here: https://github.com/neurosnap/zmx/issues/200

PR #188, opened 2026-07-08
fix: avoid statx syscall so zmx runs on pre-4.11 kernels (#186)

### LINKED ISSUE #186: `error: Unexpected` when trying to attach a zmx session on Synology Diskstation
There's not a lot more detail I can add, tbh:

```
$ ssh blob

Using terminal commands to modify system configs, execute external binary
files, add files, or install unauthorized third-party apps may lead to system
damages or unexpected behavior, or cause data loss. Make sure you are aware of
the consequences of each command and proceed at your own risk.

Warning: Data should only be stored in shared folders. Data stored elsewhere
may be deleted when the system is updated/restarted.


offby1 in 🌐 blob in ~ bsh ❯ zmx attach text
error: Unexpected

offby1 in 🌐 blob in ~ bsh ❯ uname -a
Linux blob 4.4.302+ #86009 SMP Wed Nov 26 18:19:20 CST 2025 x86_64 GNU/Linux synology_v1000_rs1221+

offby1 in 🌐 blob in ~ bsh ❯ file .local/bin/zmx
.local/bin/zmx: ELF 64-bit LSB executable, x86-64, version 1 (SYSV), statically linked, with debug_info, not stripped
```

I'm more than happy to provide any other debugging info I can get; I know, for example, that `strace` isn't present on the box, nor are most compiler toolchains. I've never cross-compiled zig before, but if you can tell this macOS user how to do it, I can build test binaries for this as well.

offbyone:
zmx crashed with `error: Unexpected` on every invocation on a Synology DiskStation (Linux 4.4.302). `error.Unexpected` is what Zig's std returns for an unrecognized errno; here the errno is ENOSYS (38) from the `statx` syscall, which only exists on Linux >= 4.11.

Two code paths issued statx via Zig std, which calls the raw statx syscall directly and does not fall back like libc does:

  - log.zig: LogSystem.init() called file.getEndPos() -> File.stat() -> statx. This runs before command parsing, so it broke *every* subcommand, not just attach.
  - socket.zig: sessionExists() called Dir.statFile() -> statx.

Both now use fd/at-based stat instead:

  - LogSystem.init() uses posix.fstat(file.handle).
  - sessionExists() uses posix.fstatatZ() + S.ISSOCK().

Because zmx links musl, these resolve to SYS_fstat / SYS_fstatat (musl on x86_64 skips statx entirely), which are available on 4.4. Verified by disassembling the release binary: no reachable statx call remains on the attach path (the only residual references are the DWARF stack-trace unwinder and sendFile fast-paths, neither of which zmx exercises).

Fixes https://github.com/neurosnap/zmx/issues/186

### DIFF
--- a/src/log.zig
+++ b/src/log.zig
@@ -1,4 +1,5 @@
 const std = @import("std");
+const posix = std.posix;
 
 pub const LogSystem = struct {
     file: ?std.fs.File = null,
@@ -22,7 +23,9 @@ pub const LogSystem = struct {
             else => return err,
         };
 
-        const end_pos = try file.getEndPos();
+        // fstat (not getEndPos) to avoid the statx syscall; see #186.
+        const st = try posix.fstat(file.handle);
+        const end_pos: u64 = @intCast(st.size);
         try file.seekTo(end_pos);
         self.current_size = end_pos;
         self.file = file;
--- a/src/socket.zig
+++ b/src/socket.zig
@@ -44,11 +44,13 @@ pub fn cleanupStaleSocket(dir: std.fs.Dir, session_name: []const u8) void {
 }
 
 pub fn sessionExists(dir: std.fs.Dir, name: []const u8) !bool {
-    const stat = dir.statFile(name) catch |err| switch (err) {
+    // fstatatZ (not statFile) to avoid the statx syscall; see #186.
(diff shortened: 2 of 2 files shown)


neurosnap:
93247e9bf93567f997f61df5aa26b1a98555d942

Issue #176, opened 2026-06-20
Build on MacOS / Darwin with nix


dwt:
Hi there,

I'm trying to package `zmx` for nixpkgs, and am currently failing miserable to get the zig build working on darwin and linux. This is probably because I do not know anything about `zig`. :)

What I have tried is listed here: https://github.com/NixOS/nixpkgs/pull/533683

What you do in the local flake seems nicer, but doesn't seem to be workable in `nixpkgs`, as I do not have access to `github:Cloudef/zig2nix` in there (if I am not mistaken).

When I try my current build, it just seems to hang in the zig compiler and produce no output, not sure at all what is going on there.

Some help would be very much appreciated.

Also I couldn't finde any GitHub Actions that build the Mac binaries - am I just blind or are there no published instructions how to reproduce the build of the Mac binaries?

adrian-gierakowski:
There is a flake for zmx here which apparently works on Darwin https://github.com/thrawny/zmx-flake/blob/2970a3a23d80214a61af247df521beb862f5a0d7/flake.nix#L47

dwt:
That actually got me on the right track, I now have a working pull request at https://github.com/NixOS/nixpkgs/pull/533683

I would very much welcome a review from @neurosnap or anyone else in this community

dwt:
If wanted, I could try to rework the flake build in this repo to also work on darwin. I won't have time for this immediately, but if wanted I can probably now provide a pull request.

dwt:
The zmx package got merged into nixpkgs, [you can follow it's progress through the branches here](https://nixpk.gs/pr-tracker.html?pr=533683)

neurosnap:
Hi, what's the status of this issue?

dwt:
zmx is successfully packaged in nixpkgs unstable, and will be part of the next stable release. I would like to follow up with a pull request to this repo that documents this installation path in addition to the others. We can close this if you like - I would appreciate a 👍 if you would take such a pull request.

theutz:
Sorry for the necropost, but I wasn't able to build v0.7.0 on Darwin using the wonderful work that you did on v0.6.0 for nixpkgs. I attempted to fix it myself, but I have 0 experience with packaging anything for Zig. It seemed like dependency called `bindings` was failing to build properly.

theutz:
I can redirect this conversation to nixpkgs if you'd prefer. But I thought I'd post it here in case anyone else stumbled through here first.

dwt:
[...]

PR #190, opened 2026-07-08
Fix darwin build

dwt:
This modifies the build so the flake builds correctly on darwin.

`nix flake show` still creates an IFD in the zig2nix flake, but that was a problem before and could be solved by setting pname and version in the package - so nothing that got worse with this patch.

I have mentioned this in #176 - which didn't get an answer, I've provided this pull request anyway int he hope that it is minimal enough to be mergeable.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -10,7 +10,7 @@
     let
       flake-utils = zig2nix.inputs.flake-utils;
     in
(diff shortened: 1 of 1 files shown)


neurosnap (approved):
Thank you for the contribution.  I don't use nix so I'm merging without validating the change.

PR #195, opened 2026-07-13
Fix nix build failure: Update zig2nix lock

thrawny:
Regenerates `build.zig.zon2json-lock` for the newer Ghostty dependency.

The previous generated lock was stale relative to `build.zig.zon`, which meant Nix builds could enter Zig's dependency fetch path and fail in the sandbox with DNS/network errors.

The lock was generated explicitly with Zig 0.15.2; zig2nix's unversioned `latest` app now uses Zig 0.16 and produces a different cache format.

## Verification

- `nix eval --raw .#packages.x86_64-linux.default.drvPath`
- `nix build .#packages.x86_64-linux.default --print-out-paths --no-link`


neurosnap (approved):
thanks!

PR #202, opened 2026-07-22
fix(send): avoid taking client leadership

### LINKED ISSUE #201: [discussion] automation input (zmx send) should not take leadership
## Context
v0.5.0 introduced client leadership: the last client to send user-input bytes becomes the leader, and on promote the daemon asks that client for its window size and resizes the PTY if `.Resize` recieved.

But today `zmx send` speaks the same `.Input` IPC tag as an attached client's keyboard. So a short-lived send connection can:
1. briefly become leader
2. get asked to report a winsize (often while already disconnecting, the PTY size is left unchanged)
3. clear the leader on close (`leader_client_fd = null`), without handing leadership back to any remaining attach client

## Proposal (for discussion)

Add an inject/bypass path that only calls `queuePtyInput` — no `setLeader`, no resize probe. 

Sketch options:

1. **New IPC tag** (e.g. `Inject = 14`), and switch `send` call over, but it's kinda breaking change for `send` command.
2. **Opt-in flag / subcommand** (`zmx send --inject` or `zmx inject`) if changing default `send` behavior considered not elegant enough.

Happy to hear which shape you prefer before a PR.

LucasIcarus:
## Summary

- Add a dedicated `.Send` IPC tag for `zmx send`.
- Queue send input without changing the active leader or requesting a resize.
- Keep `.Input` behavior unchanged for attached clients.

Fixes #201.

## Tag name

I used `.Send` for now. It maps directly to `zmx send` and keeps `.Input` for attached clients, but I'm very open to renaming it. `Inject` was the other obvious option, though it felt a little odd.

## Tests

- `zig build test`
- `bats test` (39/39)

Older daemons ignore the new tag, so existing sessions need to be restarted after upgrading for `send` to work.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@ Use spec: https://common-changelog.org/
 - `zmx run` will now detect heredocs and add the completion marker to a newline
 - Race between `zmx kill X; zmx run X`
 - Improved claim leader detection
+- `zmx send` no longer claims client leadership or triggers a resize probe
 - Improved `ctrl+\` key detection
 - Support for linux kernel < 4.11 by avoiding statx calls
 
--- a/src/ipc.zig
+++ b/src/ipc.zig
@@ -22,8 +22,9 @@ pub const Tag = enum(u8) {
     LabelSet = 15,
     LabelClear = 16,
     LabelData = 17,
+    Send = 18,
     // Non-exhaustive: this enum comes off the wire via bytesToValue and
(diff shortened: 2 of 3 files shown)


LucasIcarus:
@neurosnap  Can have a quick check when u r available  ღ( ´･ᴗ･` )

neurosnap:
This looks right to me, I'll test it locally before merging, and address any issues that come up.  Thanks!

neurosnap:
8ba312d731af09173c115a9b5ba3962c18d3a184

Thanks!

PR #203, opened 2026-07-22
Add nix installation instructions

dwt:
These are some basic nix installation instructions.

I could expand as these work on all supported nix platforms (currently mostly all Linux distributions, MacOS / Darwin, and WSL (as these are also just Linux distributions).

I could not find the sources for the website, but it also contains installation instructions. If you direct me to their source I could provide a pull request there too.

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,6 +51,24 @@ brew install neurosnap/tap/zmx
 mise use zmx
 ```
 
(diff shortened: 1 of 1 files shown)


neurosnap (approved):
This seems like a good start, thanks

PR #206, opened 2026-07-23
build: canonicalize ghostty dep URL

svmhdvn:
While here, only import build.zig.zon once in build.zig.


neurosnap (approved):
Awesome, thanks

PR #207, opened 2026-07-23
Keep `zmx list` output well-formed; enforce the label charset in the daemon

chronologos:
`zmx list` prints each session as one line of tab-separated `k=v` fields, so the format has two structural separators — tab between fields, newline between sessions — and no escaping for either. Two of the printed values are free-form and can legitimately contain those characters: `cmd` and `start_dir` are whatever the session was created with (a `-- bash -c $'…\n…'` argument, an unusually named directory). One such session makes the whole listing malformed for anything that parses it — including this repo's own bats helpers.

Separately, the label charset (`[A-Za-z0-9._-]`) is what makes labels safe to print in that format, but it was only checked in the `zmx set` CLI. The daemon's `LabelSet` handler stored whatever bytes arrived, so any other client speaking the socket protocol (a script, a different client version) could store labels that `zmx get`/`zmx list` then can't faithfully print. The check belongs where the labels are stored.

Changes:
- `writeSessionLine` folds tab/newline/CR to spaces in `start_dir` and `cmd` before printing. These are display fields, so the lossy fold costs nothing and the session still lists as one well-formed line.
- `handleLabelSet` runs the same `assertLabel` charset check the CLI does; a pair failing it is dropped with a log line while the client still gets its Ack, so the invariant holds for all clients rather than just the bundled CLI.

Tests: a unit test that separator characters in `cmd`/`start_dir` can't split the line into extra fields or rows, and a bats test that sends a `LabelSet` frame directly over the session socket (the CLI would reject the input first) and confirms the out-of-charset pair isn't stored while a valid one is.

### DIFF
--- a/src/main.zig
+++ b/src/main.zig
@@ -703,6 +703,17 @@ const Daemon = struct {
 
         var kvs = label.LabelIterator.init(labels);
         while (kvs.next()) |kv| {
+            // The CLI validates before sending, but the socket accepts messages from
+            // any writer, and stored labels are printed raw into `zmx list` lines —
+            // a value carrying a tab or newline would forge extra fields or whole
+            // rows there. Enforce the same charset the CLI does, per pair; a bad pair
+            // is dropped (logged) rather than failing the whole Ack a well-behaved
+            // client is waiting on. (`assertLabel` accepts an empty value — that's the
+            // remove form.)
+            label.assertLabel(kv.key, kv.value) catch |err| {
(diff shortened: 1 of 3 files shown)


neurosnap:
Hmm, so the original PR did have validation on the daemon side but, is this ever actually an issue or is it just theoretical?  Are you connecting to the socket directly without going through the zmx client?

neurosnap:
I like the idea of being thorough but using a non-zmx client is undefined behavior.  Closing as I don't really see this as an issue.

PR #210, opened 2026-07-26
fix: exit non-zero on invalid commands

### LINKED ISSUE #178: Invalid commands should exit non-zero
Invalid commands like `zmx foo`, `zmx completions foo`, etc. should exit non-zero.

JoeJoeflyn:
Fixes #178.

## Problem

Invalid commands silently exited 0:

- `zmx foo` (unknown command) — printed help to stdout, exited 0
- `zmx completions foo` (invalid shell) — silent, exited 0
- `zmx completions` (missing shell arg) — silent, exited 0

## Fix

All three now print `error: <msg>` to stderr and exit 1. Added a small `printError` helper following the existing `printLabelError` pattern. Valid commands (`help`, `version`, `completions bash`, `completions --help`) are unchanged.

## Tests

New `test/cli.bats` covers the three invalid cases (plus alias variants) and four valid-command regressions. Full `bats test/` suite passes (48/48: 9 new + 39 existing).

### DIFF
--- a/src/main.zig
+++ b/src/main.zig
@@ -22,7 +22,52 @@ pub const std_options: std.Options = .{
     .log_level = .debug,
 };
 
-/// This is the entry point for the CLI.
+const SessionMatch = struct {
+    name: []const u8,
+    is_prefix: bool,
+
+    fn matches(self: SessionMatch, session_name: []const u8) bool {
+        if (self.is_prefix) return std.mem.startsWith(u8, session_name, self.name);
+        return std.mem.eql(u8, session_name, self.name);
+    }
+};
+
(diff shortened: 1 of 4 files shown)


neurosnap:
I think we should audit all the commands and properly print exit codes.  Would you be willing to make that change?

JoeJoeflyn:
sure

JoeJoeflyn:
done, can u take another look

neurosnap:
Thanks!

Changes to push to upstream:
  bookmark: main [move forward from e586f625aaba to 97a1423e1ed2]

JoeJoeflyn:
tks

PR #212, opened 2026-07-27
fix: attach hang against pre-0.7.0 daemons

### LINKED ISSUE #211: `zmx attach` hangs silently when session daemon is from an older version (0.6.0 daemon + 0.7.0 client)
### Summary

After upgrading zmx 0.6.0 → 0.7.0 (Homebrew tap), `zmx attach` to any session whose host daemon is still running 0.6.0 hangs indefinitely: the client connects but renders nothing, never replays scrollback, and never returns. There is no error message. Sessions created after the upgrade (0.7.0 host) attach fine.

### Environment

- macOS Darwin 25.2.0, Apple Silicon
- zmx 0.7.0 via `neurosnap/tap` (upgraded from 0.6.0)
- Terminal: Ghostty 1.3.2-dev

### Steps to reproduce

1. Install zmx 0.6.0 and create a session: `zmx attach mysession`, detach, leave the host daemon running.
2. Upgrade to 0.7.0 (`brew upgrade zmx`).
3. Run `zmx attach mysession` with the new binary.

### Observed

- `zmx attach mysession` hangs forever with a blank screen. The log shows the connection succeeding and then nothing:

  ```
  [info] (default): socket path=$TMPDIR/zmx-501/mysession
  [info] (default): ensure session session=mysession
  [info] (default): attached session=mysession
  [info] (default): client loop fd=4
  ```

- Other commands against the old daemon behave inconsistently:
  - `zmx list` — works (though one 0.6.0 session's entry is printed malformed, missing its `name=`/`pid=` fields)
  - `zmx history mysession` — works
  - `zmx get mysession` — fails fast with a clear error: `error: session "mysession" does not support labels (daemon too old?)`
- Clients that were already attached before the upgrade keep working.

### Expected

Attach should detect the daemon/protocol version mismatch and fail fast with an actionable error (like `zmx get` already does), e.g. `error: session "mysession" was created by an older zmx; restart it to attach`. Silent hang makes it look like the session itself is wedged.

### Workaround

Kill the old host process (plain `kill <host pid>`) and recreate the session with the new binary. `zmx history` still works against the old daemon, so scrollback can be salvaged first.

silvansky:
Fixes #211

## Root cause

7381df3 (released in 0.7.0) grew the `ipc.Resize` payload from 4 to 8 bytes by adding `xpixel`/`ypixel`. Daemons on both sides of that change validate `Init`/`Resize` strictly:

```zig
if (payload.len != @sizeOf(ipc.Resize)) return;
```

so a pre-0.7.0 daemon silently drops the new client's 8-byte `Init`, never registers the client, and never replays scrollback — `zmx attach` hangs on a blank screen with no error. Everything else on the wire (`Input`, `Output`, `Detach`, `Switch`, header layout, tag values) is unchanged, so attach compatibility is fully recoverable.

## Fix

The client now sends every `Init`/`Resize` twice: the current 8-byte encoding followed by its legacy 4-byte prefix (rows+cols) — see `ipc.appendSizeMessage`. Because every daemon generation length-checks the payload, each daemon drops the encoding it doesn't expect and processes the other exactly once:

- pre-0.7.0 daemon: drops 8-byte, handles 4-byte → attach works again
- 0.7.0+ daemon: handles 8-byte, drops 4-byte → behavior unchanged, no double replay

Compared to a version probe (e.g. the `LabelGet` timeout heuristic `zmx get` uses), dual-send is deterministic: no extra round-trip, and no risk of a slow machine misclassifying a new daemon and hanging again.

The first commit fixes two unrelated darwin build errors introduced by the zig 0.16 upgrade (`mode_t` is `u16` on macOS; `wakeSignalPipe` was typed with `std.os.linux.SIG` instead of `posix.SIG`).

## Testing

- Reproduced against a real 0.6.0 daemon still running from before my brew upgrade: stock 0.7.0 client captures 9 bytes (clear-screen + reset only, blank hang); patched client receives the full ~70 KB scrollback replay and detaches cleanly.
- Verified against a freshly spawned patched daemon: single replay, no duplication (dropped 4-byte dup confirmed by the existing length check).
- Added a unit test for `appendSizeMessage` wire output; `zig build test` passes.

### DIFF
--- a/src/ipc.zig
+++ b/src/ipc.zig
@@ -112,6 +112,23 @@ pub fn appendMessage(
     }
 }
 
+/// Pre-0.7.0 daemons expect a 4-byte Init/Resize payload (rows+cols, no
+/// pixel size) and silently drop the 8-byte form, hanging `zmx attach`
+/// against a running old daemon (#211). Append both encodings: every daemon
+/// drops the length it doesn't expect and processes the other exactly once.
+pub const LEGACY_RESIZE_LEN = 4;
+
+pub fn appendSizeMessage(
+    alloc: std.mem.Allocator,
+    list: *std.ArrayList(u8),
+    tag: Tag,
+    size: Resize,
+) !void {
+    const bytes = std.mem.asBytes(&size);
+    try appendMessage(alloc, list, tag, bytes);
+    try appendMessage(alloc, list, tag, bytes[0..LEGACY_RESIZE_LEN]);
+}
+
 fn writeAll(fd: i32, data: []const u8) !void {
     var index: usize = 0;
     while (index < data.len) {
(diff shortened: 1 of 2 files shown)


neurosnap:
I apologize that we missed this when releasing v0.7.0.  We are trying to stay ABI compatible but the resize struct slipped into the release.  Unfortunately people are just going to need to wipe their sessions at this point as this release has already gone out and anyone that has upgraded will have incompatible legacy sessions.

silvansky:
Sad to hear that. I have sessions that are months long lifetime. For now I have to stay with my fork of `zmx` with the fix. Until I have to reboot, this will happen in few weeks.

neurosnap:
https://github.com/neurosnap/zmx/releases/tag/v0.7.1

Thanks!

PR #223, opened 2026-07-31
fix: build on macOS with zig 0.16.0

chancez:
`main` does not compile on macOS with zig 0.16.0. Two unrelated type errors:

- `main.zig` / `loop.zig`: `cfg.log_mode` is `u32` but `File.Permissions.fromMode` takes `mode_t`, which is `u16` here. `cfg.zig` already casts for the equivalent `Dir.Permissions` calls.
- `signal.zig`: the wake handler was declared with `std.os.linux.SIG`, which `sigaction` rejects on macOS. `posix.zig` already re-exports `SIG`.

`zig build` and `zig build test` both pass on macOS 15 (aarch64) after these.

### DIFF
--- a/src/loop.zig
+++ b/src/loop.zig
@@ -725,7 +725,7 @@ pub const Daemon = struct {
                 fba.allocator(),
                 &.{ self.cfg.log_dir, session_log_name },
             );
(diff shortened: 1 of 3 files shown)


neurosnap (approved):
ty!

Issue #222, opened 2026-07-31
Replay terminal state (title, cwd, modes) to a newly attached client


chancez:
When a client attaches, zmx replays the screen contents but not the terminal *state* that produced them. The outer terminal is left with whatever state it had, which for a freshly restored window is nothing, so features that depend on that state are broken until something in the session happens to re-emit it.

I hit this as three separate symptoms before realising they look like one gap.

### Title (OSC 2)

zmx already tracks this: the binary has `change_window_title` / `window_title` from the embedded `ghostty-vt`, including a `title length exceeds max length, truncating` warning, so it is stored. It just is not sent to a newly attached client.

With a long running command in the session, so that no shell prompt hook can interfere:

```
title while the command runs:  STICKY-TITLE
after detach and reattach:     zmx-attach
```

To be clear about that second line: `zmx-attach` there is not zmx reporting anything, it is the name of my own wrapper script that runs `exec zmx attach "$1"`. With no title set, kitty falls back to the window's process name, which happens to be that script. Any other setup would show whatever its own process is called. The point is that no title is set at all.

The session knows the title is `STICKY-TITLE` throughout. The attaching client is never told.

### Working directory (OSC 7)

Covered in #160. zmx consumes OSC 7 and does not forward it, so the outer terminal's idea of the cwd stays frozen at the directory the session was created in:

```
shell's real cwd:                  /private/tmp
kitty's recorded cwd:              ~/.dotfiles
new --cwd=current split lands in:  ~/.dotfiles
```

### Keyboard protocol and focus tracking modes

Covered in #168. If a program such as Neovim enables the kitty keyboard protocol (`CSI > 1 u`) and focus tracking (`CSI ? 1004 h`), a session switch leaves the terminal in those modes while the shell now on screen never requested them, so key release and focus reports are printed as literal text:

```
i[105;1:3u     key release report, per keypress
[I[O           focus in/out reports, on window focus alone
```

### The common shape

In each case zmx has, or could have, the state, and the attaching client ends up with a different view of it. Handling this at attach time would fix all three at once: emit the stored title as OSC 2, emit the tracked cwd as OSC 7, and reset or reapply mode state so it matches the session rather than whatever the previous occupant left behind.

Title is the one I would most like, because it cannot be worked around from outside. My current workaround is a `TRAPWINCH` in zsh that re-runs the title hooks when the wrapper signals the session shell after attaching, which works but has two limits that a zmx-side fix would not:

[...]

Issue #160, opened 2026-05-25
Let ghostty see cwd of newly attached session


SimenLK:
Hey, love the tool, btw. Been using tmux for quite some time now, but love the tool and its philosophy. Been daily driving it for around a month now.

However, using ghostty, I see that when I open a fresh window, and attach to one of my sessions, and then open a new tab, the new tab puts me in home instead of the cwd of the session. Where I'd expect ghostty to pick up on the cwd of the session I attached to.

I have configured ghostty to open tabs in dir, but not on opening new windows.

I have no idea how this is signaled or picked up, but I can understand ghostty doesn't pick up on it when you never cd, but just appear in a session.

Thanks!

joshbainbridge:
+1 for this. If Ghostty ends up supporting OSC 7 for CWD state over SSH ([discussion](https://github.com/ghostty-org/ghostty/discussions/10267)), this could also be super useful for remote sessions as well.

neurosnap:
It looks like I can just grab the cwd (which is set with OSC7) via: https://github.com/neurosnap/ghostty/blob/a746d0f7281954eb251915f4cd9fcea4924ad999/src/terminal/Terminal.zig#L2940

I'll have to test it to make sure, but should work.

chancez:
Same problem on kitty, so this is not ghostty specific: opening a new tab or split with `--cwd=current` lands in the directory the session was first started in, not where the shell actually is.

The sequence does get emitted, it just stops at zmx. From inside a session I `cd`'d and then emitted OSC 7 explicitly:

```
shell's real cwd:                 /private/tmp     (lsof on the session shell)
kitty's recorded cwd:             ~/.dotfiles      (the session's start dir)
new --cwd=current split lands in: ~/.dotfiles
```

The window title updated at the same moment, so the shell did write the sequence and zmx did process it. It just is not forwarded to the outer terminal, so kitty's idea of the cwd stays frozen at whatever it was when the session was created.

@neurosnap, your note about reading the cwd from ghostty's `Terminal.zig` looks like it would fix this. `zmx version` already reports an embedded `ghostty_vt`, so if zmx tracked the cwd from OSC 7 and re-emitted it to the outer terminal on attach and on change, terminals would pick it up without needing per-terminal support.

[...]

PR #224, opened 2026-07-31
fix: replay the window title on attach

chancez:
Related to #222, which covers title, cwd, and modes. This handles the title case only. The modes case looks fixed already by 266a733; cwd is still open and discussed in #160.

`TerminalFormatter` has no title extra and never emits OSC 0/1/2, so a client attaching to an existing session shows whatever its terminal falls back to, usually the name of the client process, even though `Terminal` already tracks the title. `serializeTerminalState` has the `*Terminal` in hand, so it can replay it alongside the state the formatter already emits.

Verified with an A/B on the same scenario, window geometry pinned so no incidental SIGWINCH could mask the result, and nvim left running so nothing else would re-emit a title:

| build | title after detach/reattach |
|---|---|
| this patch | `nvim ~/.dotfiles (master)` |
| unpatched | `attach` |

The running-program case is the one that motivated this. A shell reprints its title at the next prompt, so the window looks correct again quickly, but a long running program does not reprint, and the wrong title stays for as long as it runs. Handling it here also avoids needing a shell-side workaround, which only helps the shell that implements it and not other programs that set titles.

Two tests added: one that the title is replayed, one that no stray OSC 2 is emitted when no title is set.

Note that `main` does not currently compile on macOS with zig 0.16.0 for unrelated reasons (#223), so I verified this on top of those fixes.

### DIFF
--- a/src/util.zig
+++ b/src/util.zig
@@ -702,6 +702,16 @@ pub fn serializeTerminalState(alloc: std.mem.Allocator, term: *ghostty_vt.Termin
         return null;
     };
 
+    // The formatter has no title extra and never emits OSC 0/1/2, so the title
+    // has to be replayed separately or an attaching client shows whatever its
+    // terminal defaults to, usually the client process name. OSC 2 does not
+    // move the cursor, so this is safe to append after the content.
+    if (term.getTitle()) |title| {
+        builder.writer.print("\x1b]2;{s}\x07", .{title}) catch |err| {
+            std.log.warn("failed to format title err={s}", .{@errorName(err)});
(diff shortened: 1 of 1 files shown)


chancez:
As another follow up to this, I think it would be super great to update the title to the session whenever it changes, similar to similar to [CWD](https://github.com/neurosnap/zmx/issues/160#issuecomment-5147862112). Then it could be displayed in `zmx list` which I think would be pretty handy.

neurosnap (approved):
Excellent, thanks so much!

PR #227, opened 2026-08-03
fix(log): make PTY input logging opt-in

rartino:
In an LLM-aided security review of zmx for my own use, the logging of input keystrokes to `{log_dir}/{session}.log` was flagged, since it means tokens, passwords etc. entered during a session gets logged and left in these files even when not echoed on screen.

This PR turns off the input keystroke logging unless environment variable ZMX_LOG_INPUT is defined.

(Apologies if you prefer not getting LLM-aided PRs; if so, just close this)


neurosnap:
Thanks, gonna just comment those logs out since it really is just for local dev.

rartino:
Ok -  I got one other security-related commit from the review for cleaning up some socket permission handling: 

https://github.com/neurosnap/zmx/commit/15b780b42b3932ce2eb8536b94238d75670395d1

Do you want that as a PR?

neurosnap:
Yes please!

PR #228, opened 2026-08-03
fix(fs): make configured modes authoritative, private by default

rartino:
This changes the permission handling of sockets and directories by avoiding the influence of umask, and removes the group permissions from the default. 

The added test: `test/permissions.bats` shows the umask-related behavior that gets adjusted by this PR.

The changes to the defaults influence functionality around group shared sockets (0750 -> 0700), making zmx default closer to tmux and screen. If you don't like that and want socket directories to get group permissions by default, you might want to reject this part, but to me it seems sensible that all forms of group sharing require opt-in via ZMX_DIR_MODE.

### DIFF
--- a/src/socket.zig
+++ b/src/socket.zig
@@ -94,7 +94,7 @@ pub fn sessionExists(io: std.Io, dir: std.Io.Dir, name: []const u8) !bool {
     return true;
 }
 
-pub fn createSocket(io: std.Io, sesh: []const u8, socket_mode: u32) !lib_posix.socket_t {
+pub fn createSocket(io: std.Io, dir: std.Io.Dir, sesh: []const u8, sesh_name: []const u8, socket_mode: u32) !lib_posix.socket_t {
     // AF.UNIX: Unix domain socket for local IPC with client processes
     // SOCK.STREAM: Reliable, bidirectional communication
     // SOCK.NONBLOCK: Set socket to non-blocking
@@ -108,11 +108,11 @@ pub fn createSocket(io: std.Io, sesh: []const u8, socket_mode: u32) !lib_posix.s
     var unix_addr = try lib_posix.initUnix(sesh);
     try lib_posix.bind(fd, &unix_addr.any, unix_addr.getOsSockLen());
     try std.Io.Dir.setFilePermissions(
-        .cwd(),
+        dir,
         io,
-        sesh,
+        sesh_name,
         std.Io.File.Permissions.fromMode(@intCast(socket_mode)),
-        .{},
+        .{ .follow_symlinks = false },
     );
     try lib_posix.listen(fd, 128);
     return fd;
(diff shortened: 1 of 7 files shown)


rartino:
(Force-pushed to rebase against the latest modifications)

src/socket.zig
```diff
@@ -106,6 +107,9 @@ pub fn createSocket(sesh: []const u8) !lib_posix.socket_t {
     errdefer lib_posix.close(fd);
 
     var unix_addr = try lib_posix.initUnix(sesh);
+    // bind() creates the socket as 0777 & ~umask, so adjust umask here
+    const old_umask = cross.c.umask(@intCast((~socket_mode) & 0o777));
```
neurosnap: This `umask` trick doesn't work in Docker or root-run processes.  Could we just use `_ = cross.c.fchmod(fd, @intCast(socket_mode));` after the call to `bind()` instead?  

I confirmed it passes your tests when using docker.

There might be a microsecond delay between the socket being created and the restricted permissions being applied but that seems acceptable to me.

neurosnap (changes requested):

src/socket.zig
rartino: Sorry, `fchmod` does not seem to work for me (and I don't think it should?; fd refers to the socket object and not the filesystem entry; the documented behavior on sockets is [unspecified](https://pubs.opengroup.org/onlinepubs/9799919799/functions/fchmod.html)...) 

I implemented it in https://github.com/rartino/zmx/commit/3b9e4edebac509ce5852455047d7ca54c43154ac to try, and this is what I see:
```
$ git log -n1 --format="%h %s"
3b9e4ed Try fchmod instead of umask
$ zig build
$ bats test/permissions.bats 
permissions.bats
 ✗ permissions: session socket is private regardless of umask
   (in test file test/permissions.bats, line 16)
     `[ "$(mode_of "$ZMX_DIR/perm-default")" = "600" ]' failed
   killed session perm-default
 ✗ permissions: ZMX_DIR_MODE widens the socket for group sharing
   (in test file test/permissions.bats, line 29)
     `[ "$(mode_of "$ZMX_DIR/perm-shared")" = "660" ]' failed
   killed session perm-shared

2 tests, 2 failures
```
So I've been testing a filename based permission change in this PR (std.Io.Dir.setFilePermissions). It works in my setup, can you try it in yours?

But this implementation opens for a "what if someone swaps out the file" scenario. But at least only possible for the group members we already trust enough to give access to these sockets. The bad scenario would be if they drop a hardlink to one of our files they want to read (which usually is forbidden for normal users).

Maybe it is better to go back to just rely on the directory permission to protect the socket. That might be fine for security, but I think it does not work as intended for the actual functionality (with umask 022 you will get sockets other group members cannot connect to?)

neurosnap: Ah darn, I didn't see the "If fildes refers to a socket, the behavior of fchmod() is unspecified."

I'll test std.Io

PR #229, opened 2026-08-04
feat(run): export ZMX_TASK in task sessions

thrawny:
## Goal

Let shell configuration distinguish sessions created by `zmx run` from interactive sessions. `ZMX_SESSION` is set in both cases, so task shells currently have no reliable way to suppress prompts, greeters, direnv messages, and other interactive-only output that otherwise lands in task scrollback.

This exports `ZMX_TASK=1` in the exec child when `is_task_mode` is set and documents the environment contract. BATS coverage verifies that both `ZMX_TASK` and `ZMX_SESSION` reach the task process.

## Reviewer focus

Please focus on whether the variable is injected at the right point in the fork/exec path without affecting interactive attach sessions or the parent process.

## Deliberate scope

The variable applies when `zmx run` creates the shell. It cannot be retroactively added to a shell from an existing interactive session, and this change intentionally does not alter task-shell selection or list output.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ Use spec: https://common-changelog.org/
 - We now track cwd changes via OSC7
 - Replay window title on attach
 - `ZMX_NO_DETACH_KEY` env var to disable `ctrl+\` keybinding
+- `ZMX_TASK=1` inside sessions created by `zmx run`
 
 ### Fixed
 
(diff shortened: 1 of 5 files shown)


neurosnap:
Hi!  What happens if a `zmx` session is mixed-uses where you originally ran `zmx attach` and then switch to `zmx run` or vice-versa?

thrawny:
TBH I think I was zoomed in too hard on my own use case here so this did not really cross my mind. I basically never use zmx run on an existing session, always fresh/recreated sessions and never for anything interactive, just dev servers etc, so having those sessions clogged with my starship prompt etc was annoying me a bit.
I had this in my local fork for a while and thought I could see if it would make sense to merge it. Should have probably made an issue first now that I think of it. You live and you learn. Closing the PR, thx for looking at it :)

PR #230, opened 2026-08-04
Re-attach fixes: answer queries after detach, redraw at same size, focus events

chronologos:
Fixes for programs that keep interacting with the terminal across a detach/re-attach. Draft — the parts are independent and I'm happy to split them (e.g. 1 on its own, 3 separately, 2+4 for discussion) or drop any.

Rebased onto current `main`. The title-restore piece this PR originally carried is gone: `serializeTerminalState` now replays the title upstream, so that's covered.

**1. Answer terminal queries again after a detach.** With no client attached the daemon answers DA1/DA2 itself so a program probing the terminal isn't left waiting (this is what avoids fish's 10s startup delay). That was gated on `has_terminal_client`, which was set on the first `.Init` and never cleared — so after the first detach nothing answered anymore, and a program started or probing while detached just hung. The check is now `hasTerminalClient()`, computed from the client list, so it's right again after any detach. `handleDetachAll` now goes through `closeClient` so the leader fd is cleared there too.

**2. Same-size re-attach makes the program redraw.** After a re-attach the program should redraw, because what's on screen is now the daemon's replay rather than what it last drew. A size change already causes that via SIGWINCH, but the kernel sends no signal when the new client has the same size. In that case `handleInit` now sets rows-1, waits 5 ms, and restores (`forceSigwinchIfUnchanged`). Rows rather than cols so nothing re-wraps. I measured bash/less/vim/nano/curses on macOS: all observe both sizes at any gap ≥ 1 ms and start repainting within 0.2–1 ms; the only visible effect is one extra repaint. It is a 5 ms pause in the loop on re-attach, so I'd understand if you'd rather not have it — it's the most optional piece here.

**3. Snapshot at the client's size.** `handleInit` serialized the terminal and resized it afterwards, so a client of a different width received a snapshot that its own terminal wrapped a second time. The daemon terminal is now resized to the leader's size first, then serialized, then the PTY is resized. The old comment worried about cursor position; it stays consistent because the cursor is serialized from the same, already-resized terminal, and the shell's own SIGWINCH redraw still arrives after the snapshot since the PTY resize comes last. The prompt-redraw guard is shared with `handleResize` via `resizeTerm`.

**4. Focus in/out on attach/detach.** If the program asked for focus reports (DECSET 1004), send `CSI O` when the last attached client leaves and `CSI I` when one returns, as a terminal does when its window loses/gains focus. Programs that use focus to decide whether anyone is watching (e.g. before notifying) otherwise think they're focused while detached. Note the attaching terminal may send its own `CSI I` as well; two focus-ins are harmless but I'm open to dropping the attach half.

**Tests.** A unit test for `hasTerminalClient` (attach/detach/detach-all). `test/reattach.bats` uses a small socket client in place of a tty-backed `zmx attach` to check that DA1 is answered while detached and after a detach, is left to the client while one is attached, and that a same-size re-attach produces a SIGWINCH (verified to fail with the bounce disabled). `zig build test` and the full bats suite pass locally on Zig 0.16.

### DIFF
--- a/src/loop.zig
+++ b/src/loop.zig
@@ -241,7 +241,18 @@ fn daemonLoop(daemon: *Daemon, gpa: std.mem.Allocator, io: std.Io, server_sock_f
     var marker_carry: [32]u8 = undefined;
     var marker_carry_len: usize = 0;
 
+    var had_terminal_client = daemon.hasTerminalClient();
+
     daemon_loop: while (daemon.running) {
+        // If the program asked for focus reports (DECSET 1004), send focus-out
+        // when the last attached client leaves and focus-in when one returns,
+        // as a terminal would when its window loses/gains focus.
+        const has_terminal_client = daemon.hasTerminalClient();
+        if (has_terminal_client != had_terminal_client and term.modes.get(.focus_event)) {
+            daemon.queuePtyInput(gpa, if (has_terminal_client) "\x1b[I" else "\x1b[O");
+        }
+        had_terminal_client = has_terminal_client;
+
         poll_fds.clearRetainingCapacity();
 
         try poll_fds.append(gpa, .{
@@ -343,10 +354,9 @@ fn daemonLoop(daemon: *Daemon, gpa: std.mem.Allocator, io: std.Io, server_sock_f
                     // When no real terminal client has attached yet, respond to
                     // terminal queries (e.g. DA1/DA2) on behalf of the terminal.
                     // This prevents fish from waiting 10s for unanswered queries.
-                    // `has_terminal_client` is only set when a client sends .Init
-                    // (a real zmx attach), not when a `zmx run` tail-only client
-                    // connects.
-                    if (!daemon.has_terminal_client and
+                    // Only clients that sent .Init (a real zmx attach) count,
+                    // not a `zmx run` tail-only client.
+                    if (!daemon.hasTerminalClient() and
                         daemon.pty_write_buf.items.len < Daemon.PTY_WRITE_BUF_MAX)
                     {
                         util.respondToDeviceAttributes(gpa, &daemon.pty_write_buf, buf[0..n]);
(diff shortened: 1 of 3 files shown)


neurosnap:
Hi, please fix conflicts.  We recently merged a commit that replays window title on re-attach so that doesn't need to be included: https://github.com/neurosnap/zmx/commit/0843272cbf262c2eba8abf87e8eeacf57443db38

Everything else looks good but I need to pull it down and QA.  Thanks!

iantay-ant:
done!

chronologos:
done!

src/loop.zig
```diff
@@ -617,6 +627,41 @@ pub const Daemon = struct {
...
+        defer term.flags.shell_redraws_prompt = saved;
+        try term.resize(gpa, .{ .cols = cols, .rows = rows });
+    }
+
+    /// If the PTY is already at `ws`, briefly shrink it by one row so the
+    /// program receives a SIGWINCH and redraws (setting an identical size
+    /// raises no signal). Rows rather than cols so nothing re-wraps. The pause
+    /// lets the program read the intermediate size before the caller restores
+    /// it; programs react in well under 1ms, 5ms is margin.
+    fn forceSigwinchIfUnchanged(io: std.Io, pty_fd: i32, ws: cross.c.struct_winsize) void {
```
neurosnap, line 646: I don't fully understand why we need this, what happens when we don't call this fn?

chronologos: without this, programs that repaint incrementally (e.g. Ink-based TUIs, Claude Code) don't know to repaint and sometimes they will just show stale screen content until the next real resize or Ctrl-L.

neurosnap: Gotcha, it seems like we could just send a `SIGWINCH` directly to the process group which would bypass `ioctl` no-op when the dimensions are identical.

I'm going to try that first before relying on a pixel-shift which feels a little jank.

neurosnap:
Thanks!

Changes to push to upstream:
  bookmark: main [move forward from fdc244908886 to e586f625aaba]

PR #231, opened 2026-08-05
feat: new switch command

neurosnap:
This is an explicit command to switch to a session from within a session

BREAKING CHANGE: we now block users when trying to switch to a session from within a session using `zmx attach`


maxandersen:
will zmx switch work when not inside session ?

neurosnap:
I don't think this is going to make it, I linked the issue where we are trying to figure out what to do.

PR #232, opened 2026-08-05
fix: drop stray NUL and decode the OSC 7 cwd

chancez:
Related to #222. Fixes the two cwd bugs left over from that thread, one commit each.

The NUL one is no longer cosmetic, which is why it is here rather than filed separately. My kitty writes a session file so it can restore my zmx sessions on login, and it records the cwd it was told. It persisted the NUL, then failed to parse its own session file on the next start, so kitty would not open at all.

### Stray NUL in the replayed OSC 7

The formatter's `extra.pwd` writes `terminal.pwd.items`, and `Terminal.setPwd` appends a NUL sentinel to that buffer. `getPwd()` returns the same bytes without it.

Verified with a control, capturing the replay through a pty. Redirecting stdout defeats the replay and `script` buffers until exit, so both of those measure nothing:

| build | last OSC 7 in the replay | NUL count |
|---|---|---|
| pre-fix | `\x1b]7;file://HOST/tmp/zmx%20replay%20dir\x00\x1b\\` | 1 |
| this patch | `\x1b]7;file://HOST/tmp/zmx%20replay%20dir\x1b\\` | 0 |

`serializeTerminal` had the same bug, so `zmx history --vt` was emitting it too. Both are fixed.

### cwd stored as a raw URI

`zmx list` printed `cwd=file://HOST/tmp/zmx%20spaced%20dir`, and the undecoded escaping broke the chdir in `Daemon.run`, so the new session silently landed in the daemon's inherited cwd instead. Decoding once at `setCwd` fixes both.

The host still decides whether we chdir, per your note about SSH boundaries. A remote cwd is reported but never entered, and is left out of the `Switch` payload rather than handed to a new session that could not use it.

One thing beyond what I described in the issue: that check used `mem.eql` against `gethostname()`, so a missing host, `localhost`, and short-name/FQDN mismatches all read as remote and disabled the chdir even locally. Handled now.

`ipc.Info` is explicitly frozen, so the host is not exposed in `zmx list`; the daemon keeps `cwd_is_local` internally. If you want `list` to distinguish a remote cwd, that needs a new `Tag`, which felt like a separate change.

### Testing

Unit tests for the NUL (including the `--vt` history path), a table test for `parseOsc7Cwd`, and `test/cwd.bats` covering `list` output and the chdir end to end. 77 unit tests and the full 42-test bats suite pass.

Switching was verified in a throwaway kitty instance rather than my own terminal, since `zmx attach` from inside a session retargets the existing client. The switch carried the decoded path and `lsof -d cwd` confirmed the new daemon started in it.

Unrelated, but it affects anyone reproducing this locally: `Dockerfile` hardcodes the x86_64 zig tarball, so on Apple silicon the image runs under Rosetta and `zig build` dies with `error: rosetta error: bss_size overflow`, making `zmx run integration` unusable. I ran bats in an otherwise identical image built with the aarch64 tarball. Happy to send that as its own PR.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ Use spec: https://common-changelog.org/
 ### Fixed
 
 - Clear screen when switching sessions to prevent term state corruption
+- Stray NUL byte in the OSC 7 sequence replayed on attach
 
 ### Changed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,8 @@ Use spec: https://common-changelog.org/
 
 - Clear screen when switching sessions to prevent term state corruption
 - Stray NUL byte in the OSC 7 sequence replayed on attach
+- The OSC 7 cwd is now decoded before the chdir, so a new session can start in
+  a directory whose name needed percent-encoding
 
 ### Changed
 
--- a/src/loop.zig
+++ b/src/loop.zig
@@ -560,7 +560,18 @@ pub const Daemon = struct {
     running: bool = true,
     pid: i32 = undefined,
     command: ?[]const []const u8 = null,
+    /// The session's working directory in OSC 7 form, `file://<host><path>`.
+    /// Kept as a URI rather than a path so `zmx list` shows the host, which is
+    /// what tells you a session is inside SSH. Points into `cwd_buf` once set,
+    /// so a Daemon must not be copied by value after that.
     cwd: []const u8 = "",
+    /// The same directory as a path that can be opened: percent-decoding
+    /// applied, scheme and host stripped. Empty when the cwd is on another
+    /// host, since then it names no directory here and nothing should chdir
+    /// into it. Points into `cwd_path_buf`.
+    cwd_path: []const u8 = "",
+    cwd_buf: [std.fs.max_path_bytes]u8 = undefined,
+    cwd_path_buf: [std.fs.max_path_bytes]u8 = undefined,
     has_pty_output: bool = false,
     has_had_client: bool = false,
     has_terminal_client: bool = false, // true only after a real attach (.Init received)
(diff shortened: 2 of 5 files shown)


neurosnap:
Hmm what if we keep the cwd stored as the std.Uri (maybe just remove the sentinel) and then parse it correctly for `chdir`?  I kind of like that `zmx list` shows the host in the std.Uri format unless you feel really strongly that will be annoying for users.  It's just nice to see the host because it signals you are in an SSH session.

chancez:
Good call, done. Storing the URI and parsing only for chdir is a better split, and I agree the host is worth showing.

`Daemon.cwd` now keeps the OSC 7 form and a second field, `cwd_path`, holds the decoded path used for the chdir. `cwd_path` is empty when the host is not this machine, which is also what keeps a remote cwd out of the `Switch` payload, so that no longer needs a separate `is_local` flag on the daemon.

A caller that only has a local path (`zmx run`, `zmx attach`) gets a URI built by a new `util.toOsc7Cwd`, so `cwd` has one shape regardless of source rather than being a URI only after the shell reports one. That also means `zmx list` shows the host for a plain local session, not just an SSH one.

`zmx list` output now:

```
name=local1   ... cwd=file://CHANCEZ-M-2YPG/private/tmp/zmx%20spaced%20dir
name=remote1  ... cwd=file://prod-box/srv/app
```

The percent-encoding is still there, since the value stays a URI. If you would rather `list` decode just for display while keeping the stored value a URI, that is a small change on top, but I left it as-is so what you see is what the shell reported.

Verified the remote case does not chdir (`set cwd=file://prod-box/srv/app path=` in the log, with the daemon still in its inherited cwd) and the local spaced-dir case still does. Also added a round-trip test, `toOsc7Cwd` then `parseOsc7Cwd`, covering spaces, `%`, and quotes.

neurosnap:
https://erock-public-ci-zmx.pgs.sh/1786459370/index.html

neurosnap:
bookmark: main [move forward from b7033a8ee89b to 7a001c42b49f]

PR #237, opened 2026-08-14
fix: allow `*` to match all sessions

NgoQuocViet2001:
## Summary

- allow a bare `*` wildcard to create an empty-prefix matcher
- keep scoped wildcards and `ZMX_SESSION_PREFIX` validation unchanged
- add regression coverage for matching arbitrary session names

The changelog documents `zmx kill "*"` as the way to target every session, but the parser currently strips `*` and rejects the resulting empty name with `SessionNameRequired`.

## Test plan

- Ran: `env -u ZMX_SESSION_PREFIX zig test src/socket.zig -lc` (7 tests passed)
- Ran: `git diff --check`


neurosnap (approved):

neurosnap:
https://erock-public-ci-zmx.pgs.sh/1786710751/

PR #238, opened 2026-08-14
fix: write/reset osc 10/11/12 on attach/detach

dmun:
Fixes restoration of OSC 10/11/12 colors on reattach.

Also changed the reset sequence to include the OSC 10/11/12 resets. Kitty does this for `\x1bc`, but terminals like Ghostty don't (which I'm using in the example).

## Before
https://github.com/user-attachments/assets/90094f63-cd48-43b2-bb64-af369ec6d5e0

## After
https://github.com/user-attachments/assets/8ec86ad7-2362-437d-9d11-241e3e2218ee

### DIFF
--- a/src/main.zig
+++ b/src/main.zig
@@ -1312,12 +1312,14 @@ fn attach(gpa: std.mem.Allocator, io: std.Io, daemon: *Daemon) !void {
     var orig_termios: cross.c.termios = undefined;
     const stdin_is_tty = cross.c.tcgetattr(lib_posix.STDIN_FILENO, &orig_termios) == 0;
 
+    // RIS, OSC 10/11/12
+    const restore_seq = "\x1bc\x1b]110\x1b\\\x1b]111\x1b\\\x1b]112\x1b\\";
+
     defer {
         if (stdin_is_tty) {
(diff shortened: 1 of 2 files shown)


src/main.zig
```diff
@@ -1312,12 +1312,14 @@ fn attach(gpa: std.mem.Allocator, io: std.Io, daemon: *Daemon) !void {
     var orig_termios: cross.c.termios = undefined;
     const stdin_is_tty = cross.c.tcgetattr(lib_posix.STDIN_FILENO, &orig_termios) == 0;
 
+    // RIS, OSC 10/11/12
+    const restore_seq = "\x1bc\x1b]110\x1b\\\x1b]111\x1b\\\x1b]112\x1b\\";
```
neurosnap, line 1316: I'm curious, is there a reason why you moved this out of the defer/switch branches?

dmun: I assumed that they always have the same value so I thought I'd just make a variable for them

neurosnap (approved):

neurosnap:
Thanks!

Changes to push to upstream:
  bookmark: main [move forward from a959c5a624ef to 883b02d7dec8]

Issue #151, opened 2026-05-05
Can not create a new session with a command when ZMX_SESSION is set


z4nsh1n:
The following shell script is based on tmux-sessionizer:
#!/bin/sh 

if [[ $# -eq 1 ]]; then
    selected=$1
else
    #selected=$(find ~/code ~/Downloads/ -mindepth 1 -maxdepth 3 -type d | fzf)
    selected=$(fd . ~ ~/code ~/.config ~/Downloads/ ~/Music/ --maxdepth 2 -t d| fzf)
  
fi

if [[ -z $selected ]]; then
    exit 0
fi

selected_name=$(basename "$selected" | tr . _)

# str="cd $selected;fish"
zmx a $selected_name sh -c "cd $selected;fish"
# zmx a $selected_name sh -c "cd $selected;ZMX_SESSION=\"\" fish -C \"set ZMX_SESSION=$selected\""

First time I run this it works. When in a spawned session zmx does not use the command but spawns the default shell.
If I clear ZMX_SESSION the shellscript works but the session gets locked up.

I would like to execute this in zmx sessions to spawn new sessions

Jelloeater:
I'm seeing a similar issue...
When I start a new WezTerm session from inside a JetBrains ZMX attached session, the env var carries over and doesn't get unset... weird.

Caused me to not have auto attach setup for my IDE. Odd...

Ex

```shell
alias wez-window='wezterm cli spawn --new-window --cwd .'
alias wez-tab='wezterm cli spawn --cwd .'
alias t='(&>/dev/null nohup wezterm start --always-new-process --cwd . &)'
alias tt='(&>/dev/null nohup wezterm start --always-new-process --cwd . zmx attach "$(basename $PWD)-$(date +%s)" &)'

#############################

export PATH="/Users/jesse/.local/bin:$PATH"

if [[ "$TERM_PROGRAM" == *code* ]]; then
    export IDE_TERM='VS'
fi
if [[ "$TERMINAL_EMULATOR" == *JetBrains* ]]; then
    export IDE_TERM='JB'
fi
if [[ "$TERM_PROGRAM" == *WezTerm* ]]; then
    unset IDE_TERM
fi

# Auto Attach might cause issues
# ??TODO?? Why does ZMX_SESSION not get unset when spawning a new session, yeah, it's a env-var to carry, but it shoud NOT persist when spawning a new terminal
 
# if [[ "$IDE_TERM" ]]; then
[...]

Issue #221, opened 2026-07-31
Expose which session a client is attached to (client pid in `zmx list`, or `zmx whoami`)


chancez:
Given the process id of a `zmx attach` client, there is no supported way to ask which session that client is currently displaying.

`zmx list` reports `clients=1` as a count only, never an identity, and `zmx get` returns just labels. The client's own argv is not reliable either, because of the session-switch behavior discussed in #151 and #167: after a switch, argv still names the session the client was launched with, while it is connected to a different one.

```
client pid 11533   argv:   zmx attach ev.host
                   socket: /tmp/zmx-501/ev.other
```

The only approach I found is matching unix socket peer addresses with `lsof`, which needs two invocations because the client holds an unnamed peer address and the daemon side holds the path:

```sh
lsof -p "$PID" -a -U     # -> ->0xb831feaf3cbe7349
lsof -U | grep "$PEER"   # -> /tmp/zmx-501/ev.other
```

That depends on undocumented socket layout, costs around 50ms, and is easy to get subtly wrong: while working on this I had a stale peer address resolve to the wrong session, which is exactly the kind of mistake that becomes destructive in the use case below.

### Why I need it

I run kitty with each window's process being `zmx attach <name>`, so that quitting kitty leaves the sessions running and restoring the saved layout reattaches to them. Two things in that integration need to map a window back to a session:

1. **Closing a window should kill its session**, otherwise closing a tab silently orphans one. Resolving the name from launch arguments killed the *wrong* session for me after a switch, destroying the work running in it.

2. **Opening a new split should inherit the session's current directory.** That is a two step lookup: window to session name, then session name to cwd. The second step is #160, and it needs the first step to even begin.

Both steps are currently `lsof` calls against undocumented internals.

### What would help

Anything that exposes the mapping, for example:

- a `client_pids=` field in `zmx list`
- a `zmx whoami [--pid N]` that prints the session name for a client

Combined with the JSON output in #220 this would make terminal integrations straightforward. Worth noting that if OSC 7 forwarding lands for #160, the second lookup goes away on its own, and this would be the only remaining hole.

neurosnap:
My gut reaction is to replace clients=N with a list of pids in zmx list.

chancez:
Yah that feels natural to me.

PR #240, opened 2026-08-16
feat(attach): add -n to attach without switching sessions

NgoQuocViet2001:
## Summary

Implements the `zmx a -n {sesh}` flag discussed in #151: attach in the current terminal instead of switching the session named by `ZMX_SESSION`.

- add `-n`/`--new` to `attach`
- only read flags before the session name, so an `-n` inside the session command is still passed through
- document the switch behavior of `ZMX_SESSION` in `zmx help`
- add fish and nu completions for the flag

## Why

`attach` switches the calling terminal when `ZMX_SESSION` is set. A non-interactive caller that merely inherited the variable has no way to know that, so it retargets a terminal it never meant to touch. #151 reports the caller's session being left with no client, and a launcher that maps windows back to sessions by argv then killing the wrong one.

`zmx run` covers background commands, but not the case where the caller does want an interactive client and simply inherited `ZMX_SESSION` from its parent.

## Behavior

With `ZMX_SESSION=host` set:

```
zmx attach target        before: host        after: (none)
zmx attach -n target     before: host        after: host target
```

Without `ZMX_SESSION`, and for every existing invocation, behavior is unchanged.

## Test plan

- Ran: `zig build test` (green; new cases fail without the fix)
- Ran: `zig build`
- Checked: the A/B above against a built binary with an isolated `ZMX_DIR`
- Checked: `zmx attach dev vim -n` still passes `-n` to the command

### DIFF
--- a/src/completions.zig
+++ b/src/completions.zig
@@ -160,6 +160,7 @@ const fish_completions =
     \\complete -c zmx -n "__fish_is_nth_token 2; and __fish_seen_subcommand_from c completions" -a 'bash zsh fish nu' -d Shell
     \\
     \\# Subcommand flags
+    \\complete -c zmx -n "__fish_seen_subcommand_from a attach" -s n -l new -d 'Attach here instead of switching the ZMX_SESSION terminal'
     \\complete -c zmx -n "__fish_seen_subcommand_from r run" -s d -d 'Detach from the calling terminal; use `wait` to track its status'
     \\complete -c zmx -n "__fish_seen_subcommand_from r run" -l fish -d 'Required when the session runs fish shell'
     \\complete -c zmx -n "__fish_seen_subcommand_from l list" -l short -d 'Short output'
@@ -180,6 +181,8 @@ const nu_completions =
     \\
     \\export extern "zmx attach" [
     \\    name: string@"nu-complete zmx sessions"
+    \\    -n
+    \\    --new
     \\    ...rest: string
     \\]
     \\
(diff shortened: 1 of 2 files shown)


neurosnap:
Yeah I'm not really sure if we can get this to work well at all.  Using your PR:

zmx a test.1 -> zmx a -n test.2 -> test.1 is unrecoverable

We might just need to add a note in the README that if ZMX_SESSION is inherited in a child shell then they need to unset it before trying to attach.

Nested sessions is just not supported and I don't see a clear way to enable support with undesirable behavior.

NgoQuocViet2001:
You're right, and I should have caught this before opening it.

I went back to `attach()` to see why your repro breaks. With `-n` the client skips `switchSesh` and falls straight through to `sessionConnect` → `tcsetattr(STDIN_FILENO, raw)` → `clientLoop`, all on the terminal the outer client already owns. So the flag never makes a nested *view* — it makes a second client on one TTY, with both loops reading the same stdin. `test.1` is unrecoverable because its client is still there fighting for input, not because of anything the switch did.

That is the nesting problem itself, and a flag on `attach` cannot reach it. Closing this.

Opened #242 with the README note instead — a `## nested sessions` section saying `attach` reads `ZMX_SESSION` and switches the caller, that `env -u ZMX_SESSION` is the fix when the variable was inherited rather than chosen, and that `zmx run` is the right tool for non-interactive callers.

Two follow-ups, both independent of this:

- #241 (`--labels`) was stacked on this branch. I rebased it onto `main`, so it stands alone now and nothing here blocks it.
- The destructive half of #151 — a launcher mapping windows back to sessions by argv and killing the wrong one after a switch — is untouched by the docs note. That needs the client to be able to report which session it is actually showing, which is @chancez's #221. Happy to take that one if you want it, but I won't start without your say-so given how this went.

Issue #200, opened 2026-07-21
feature: sticky labels


neurosnap:
When creating a session within a session it might be desirable to copy-over all the labels from the parent session to the child.

Here's one idea:

```bash
zmx attach parent
zmx attach --labels "$(zmx get .)" child
```

Idea from: https://github.com/neurosnap/zmx/pull/185

maxandersen:
I think that should be doable but with this it requires shell expansion - where as `zmx attach --labels-from=. child` would allow it to be done without going through a shell.

I still would love option to enable auto stickyness but i guess thats another issue :)

ruudk:
Another angle on the same flag: I use zmx as a process supervisor. A control
plane starts many git worktrees, each with roughly 15 processes, all in their
own session. Every one of them wants the same handful of labels — project,
worktree, process name — so the supervisor can later find and tear down exactly
what it owns.

Today that means a second round-trip per session right after creating it. And if
the supervisor dies in between, the session is left running with no labels at
all, so nothing can attribute it afterwards.

So `--labels` at creation would help here too, not just for inheriting a parent
session's labels.

chancez:
For what it's worth I'm sorta doing this by having my shell automatically set various labels automatically as things change. Eg: I have a set of shell hooks to set my cwd on my current session, as well as a prog label for the currently running program (if any). I hadn't considered inheriting labels, but that's a cool idea I might have to explore. That could even lend itself to an almost hierarchical structure which could be useful too.

PR #241, opened 2026-08-16
feat(attach): apply labels as the session is created

NgoQuocViet2001:
Implements `--labels` from #200, in the form you sketched there.

```
zmx attach --labels "project=api role=worker" build
```

The value takes the same `key=value ...` string `zmx set` does, so `zmx attach --labels "$(zmx get .)" child` gives the parent-inheritance case from the issue without new syntax. `--labels=...` works too.

@ruudk's supervisor case in the thread is the sharper motivation: creating and then labelling is two round trips, and if the supervisor dies in between, the session runs on with no labels and nothing can attribute it. Applying them at creation closes that window.

Labels are validated **before** `ensureSession`, so a rejected label does not leave behind a session the caller never asked for:

```
$ zmx attach --labels "name=reserved" build
error: "name" is a read-only built-in field
$ zmx list --short
(nothing)
```

The validation loop moved out of `labelSet` into `assertLabels`; `labelSet` still calls it, so `zmx set` is unchanged.

Not doing `--labels-from=.` from @maxandersen's comment here — it is a separate resolution step and I did not want to guess the shape. Happy to add it if you want it in the same change.

**Now standalone.** This was stacked on #240; after your comment there I rebased it onto `main` so it does not depend on that flag landing. One commit, no `-n` in it.

### Test plan

- Ran: `zig build test` (4 new cases: both `--labels` spellings, combined with a session command, missing value, and flag absent)
- Ran: `zig build`
- Checked: `attach --labels "project=api role=worker" build` then `zmx get build` → `project=api role=worker`
- Checked: `bad!=x`, `name=reserved` and `=novalue` each print their existing error and leave `zmx list` empty
- Checked: `--labels` with no value exits 1 with a message instead of a stack trace
- Checked: `zmx attach dev vim --labels` still passes `--labels` to the command

### DIFF
--- a/src/completions.zig
+++ b/src/completions.zig
@@ -160,6 +160,7 @@ const fish_completions =
     \\complete -c zmx -n "__fish_is_nth_token 2; and __fish_seen_subcommand_from c completions" -a 'bash zsh fish nu' -d Shell
     \\
     \\# Subcommand flags
+    \\complete -c zmx -n "__fish_seen_subcommand_from a attach" -l labels -d 'Apply "key=value ..." labels as the session is created' -r
     \\complete -c zmx -n "__fish_seen_subcommand_from r run" -s d -d 'Detach from the calling terminal; use `wait` to track its status'
     \\complete -c zmx -n "__fish_seen_subcommand_from r run" -l fish -d 'Required when the session runs fish shell'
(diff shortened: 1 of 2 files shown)


neurosnap (approved):
Excellent job!

Changes to push to upstream:
  bookmark: main [move forward from 9575587bbe36 to 3604a1770c22]

PR #242, opened 2026-08-17
docs: note that nested sessions are not supported

NgoQuocViet2001:
The README note you suggested on #240, since the flag there does not work.

Adds a short `## nested sessions` section before `## shell prompt`: `attach` reads `ZMX_SESSION` and switches the calling terminal, which is surprising when the variable was inherited rather than chosen, and the fix is `env -u ZMX_SESSION` (or `zmx run` for non-interactive callers).

Closing #240 in favour of this — details in that thread.

### DIFF
--- a/README.md
+++ b/README.md
@@ -121,6 +121,25 @@ Commands:
   [h]elp                                   Show this help
 ```
 
+## nested sessions
+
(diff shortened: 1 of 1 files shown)


neurosnap (approved):
ty!

Issue #225, opened 2026-08-01
Update per-client environment variables on attach (like tmux update-environment)


chancez:
First off, zmx has been great. I'm daily driving it with kitty and it's working really well, so everything below is me iterating on my own setup rather than hitting a wall. Thanks for building it.

A session's environment is captured when it's created and never updated, so any variable that describes the client that attached is wrong for every client after the first. tmux handles this with `update-environment`, and its default list is the same kind of variable: `DISPLAY`, `SSH_AUTH_SOCK`, `SSH_AGENT_PID`, `SSH_CONNECTION`, `WINDOWID`, `XAUTHORITY`.

`SSH_AUTH_SOCK` is probably the most common case. Attach from a second ssh connection and agent forwarding in that session still points at the first connection's socket, which is gone.

My case is kitty. It exports a socket path, its pid, and a window id, and quitting kitty invalidates all three:

```
session env : KITTY_LISTEN_ON=unix:/tmp/kitty-4265  KITTY_PID=4265  KITTY_WINDOW_ID=10
reality     : /tmp/kitty-10597                      10597           4
```

So kitty's remote control stops working in a restored session:

```
$ kitten @ ls
Error: Failed to connect to unix:/tmp/kitty-4265 with error: dial unix /tmp/kitty-4265: connect: no such file or directory
```

That's the mechanism kittens, notifications, and anything scripted against kitty's API use to reach the terminal, so a fair amount stops working at once. Any terminal that advertises itself through the environment would have the same problem.

This can't be fixed from outside the session, since nothing can change the environment of a shell that's already running. What tmux does is record and report, and let the shell opt in:

- `update-environment` lists the variables to capture from each attaching client
- `show-environment -s` prints them as shell commands, which a `precmd` hook evaluates

The recording half is the part only zmx can do. A minimal version would be enough for me:

```sh
zmx get-env <session>     # what the most recent client had, as KEY=VALUE
```

A fixed list would be fine rather than a configurable one, since the shell side can filter.

If that's more surface than you want, a hook that runs a command on attach would let me do the refresh myself without zmx knowing anything about environments, and it's a more general primitive:

```sh
zmx attach dev --on-attach 'refresh-env'   # or a config option
```

Right now I work around these one at a time by resolving the socket when I use it and matching windows by session name instead of window id. That works, but it has to be redone for every feature that reads one of these variables.

[...]

Issue #104, opened 2026-03-24
SSH agent forwarding doesn't work after disconnecting SSH and then reconnecting and attaching to the persisted `zmx` session


senderista:
I use SSH forwarding to authenticate to GitHub from my remote dev box. I have never had any issues with my current setup (using `AddKeysToAgent yes`, `ForwardAgent yes` in my `.ssh/config`) until I tried accessing GitHub from inside a `zmx` session. It worked fine until I disconnected from the remote server, reconnected, and then re-attached to the `zmx` session, after which GitHub auth fails. Is agent forwarding supposed to be handled transparently by `zmx`, or is there something I need to change in my config to make this work?

senderista:
Here is a discussion of the key forwarding issue in the context of `tmux`:

https://blogsystem5.substack.com/p/ssh-agent-forwarding-and-tmux-done

I should be able to work around this issue for now using one of the solutions in the article, but it would be really nice for this to be supported in `zmx` since it's a core part of my workflow (and anyone else who uses GitHub or other services requiring SSH auth from within a persistent session).

neurosnap:
Thanks for submitting this issue.  Looking at prior art ([shpool](https://github.com/shell-pool/shpool/blob/5c0038ce22dea27b4b823c2255209333eb3c4862/libshpool/src/config.rs#L191)), it looks like we might be able to create a symlink to the current `SSH_AUTH_SOCK` (e.g. `/tmp/ssh-auth.sock`) and then change the `SSH_AUTH_SOCK` to that tmp file.

senderista:
I guess a caveat to this approach was already in the article I linked:

> It’s insufficient. Some of the solutions linked to above leverage a symlink to keep the socket’s name stable. Those solutions refresh the symlink every time you log into the remote machine and then point SSH_AUTH_SOCK to the symlink. This is good because it solves the problem of having to monkey-patch all possible long-lived processes. Unfortunately, this breaks as soon as the latest connection to the server disconnects: once that happens, the symlink will be dangling until you establish a new session to update it.

IOW, if a long-running process started in a `zmx` session tries to do some SSH operation after you've disconnected from SSH but before you've reconnected from SSH, the symlink will be referencing a deleted socket file (because the socket file only exists as long as an SSH session exists).

senderista:
[...]

PR #245, opened 2026-08-20
feat: leader sets tracked env vars

neurosnap:
When a client assumes the leader for a zmx session it sets the tracked environment variables for the session which can the user can then print using `zmx print-env {sesh}`.

This can be used inside a shell precmd hook to automatically set the leader's env vars to the shell session: `eval "$(zmx print-env .)"`.

Link: https://github.com/neurosnap/zmx/issues/225
Link: https://github.com/neurosnap/zmx/issues/104


neurosnap:
Changes to push to upstream:
  bookmark: main [move forward from fb1b6b66476f to fdc244908886]

PR #251, opened 2026-08-27
feat(lib): add dependency-free zmx module

alleneubank:
## Summary

- Add a reusable `zmx` Zig module for embedders.
- Add a compile-only dependency oracle for the public library surface.
- Keep the dependency-free surface independent of CLI `build_options` and
  `ghostty-vt` sources.
- Preserve existing CLI runtime behavior.

## Motivation

Embedding zmx currently requires consumers to reach through the executable's
source layout. A first-class module gives embedders a supported library entry
point while keeping CLI-only configuration and terminal-rendering dependencies
out of the dependency-free surface.

## Verification

- [x] `zig build`
- [x] `zig build test`
- [x] `zig build check-lib-deps`
- [x] `zig fmt --check src/ build.zig`
- [x] `nix shell nixpkgs#bats -c bats test/` — 44 tests passed

### DIFF
--- a/build.zig
+++ b/build.zig
@@ -31,6 +31,16 @@ pub fn build(b: *std.Build) void {
     });
     exe_mod.addOptions("build_options", options);
 
+    // Public library module exposed to package consumers. It intentionally
+    // has neither build_options nor ghostty-vt so embedders can link the zmx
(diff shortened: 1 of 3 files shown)


alleneubank:
Wrong repo! My bad.

PR #254, opened 2026-09-05
fix: let the line limit control scrollback retention

chamard:
Following up on [my comment](https://github.com/neurosnap/zmx/commit/8a6d0381a49646f0db9989b24367e315de5ccf49#r199174164): libghostty’s default 10 KB byte limit still applies after the 2,000-line limit, causing history to be pruned early (around 820 rows at 80×24).

This sets `max_scrollback_bytes = null` so only the line limit controls retention.

### DIFF
--- a/src/loop.zig
+++ b/src/loop.zig
@@ -214,6 +214,15 @@ pub fn clientLoop(client_sock_fd: i32, env_str: []const u8) !ClientResult {
     }
 }
 
+fn initTerminal(gpa: std.mem.Allocator, io: std.Io, size: ipc.Resize, cfg: *const Cfg) !ghostty_vt.Terminal {
+    return ghostty_vt.Terminal.init(io, gpa, .{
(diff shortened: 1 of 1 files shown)


src/loop.zig
```diff
@@ -214,6 +214,15 @@ pub fn clientLoop(client_sock_fd: i32, env_str: []const u8) !ClientResult {
     }
 }
 
+fn initTerminal(gpa: std.mem.Allocator, io: std.Io, size: ipc.Resize, cfg: *const Cfg) !ghostty_vt.Terminal {
+    return ghostty_vt.Terminal.init(io, gpa, .{
+        .cols = size.cols,
+        .rows = size.rows,
+        .max_scrollback_lines = cfg.max_scrollback_lines,
```
chamard, line 221: It could also be useful to make `max_scrollback_lines` configurable from an environment variable

Issue #244, opened 2026-08-19
Consider adding Structured command output retrieval like backscroll


dwt:
Backscroll https://github.com/soren-achebe/backscroll works very similar to zmx. It provides structured past command outputs.

That would be super useful for integration into coding agents, as it would allow it to iteratively discover past commands in the session. 

I imagine a simpler design than backscroll, as zmx has no need for persistence.

I also really like the way they support all the shells, and can make it even better with shell integration snippets.

What do you think?

neurosnap:
Yes I would consider it. I believe @deevus has a prototype somewhere for zmx

PR #255, opened 2026-09-06
feat(history): capture the active screen with bounded scrollback

maddada:
Hey Neurosnap!
Hope you're doing great man!

I'm the greatest fan of zmx, it's honestly such an amazing utility.

My whole ADE is based on it basically and it solved so many issues for me (Client <> Host architecture. Web. Mobile. Converting agent CLIs to a chat GUI view. Way more.)

Now I was looking into the polling being done by the Chat View with Astra and it says that it can be optimized in zmx so that the history command doesn't send rows that aren't required (like when using the tail argument)

The idea of this PR is to only get the needed rows from zmx to lower the cpu/memory required for reading from zmx (I need to poll the terminal a lot while the agent is working)

---

Summary by Astra high:

Consumers that poll a terminal's current state currently have to request all history and discard most of it. This adds daemon-side selection so the formatter and socket only handle the requested rows.

- `zmx history --screen SESSION` captures the current grid of the active screen.
- `zmx history --scrollback N SESSION` includes that grid plus up to N preceding physical rows.
- Both support plain text, `--vt`, and `--html`, with help, README documentation, and shell completions.

The existing history command and History IPC payload remain unchanged. A new optional Capture tag carries the format and scrollback count as a packed struct. An Info request follows it as an ordering barrier: an older daemon answers Info without Capture, and the CLI reports that the daemon is too old (same shape as the labels error) instead of silently returning an unbounded result or waiting for a capability timeout. Existing sessions do not need to restart to continue using ordinary history.

Open question: would you prefer extending the History payload from 1 to 5 bytes instead? Old daemons already read only the first byte and would return full history, which drops the new tag and the barrier at the cost of a silent fallback.

Selection reads Ghostty's already-updated active buffer, so cursor movement and in-place redraws are preserved. Alternate screens return only their grid. The row count refers to physical rows, including wrapping and blank rows, rather than a client's scrolled-back viewport.

Validation on macOS arm64 with Zig 0.16.0 (using a local SDK header overlay for the installed SDK):

- `zig build -Doptimize=ReleaseSafe` passed.
- `zig build test -Doptimize=ReleaseSafe` passed, 98/98 tests (three new `serializeTerminalRange` tests: screen only, requested rows with clamping, alternate screen).
- `zig fmt --check` and `git diff --check` passed.
- Disposable-session checks verified exact recent-history suffixes, the requested extra row count, alternate-screen isolation, and in-place redraws. One 1,207-line capture shrank from 11,200 bytes to 4,817 bytes with `--scrollback 512`.

Draft for review of the CLI and additive protocol shape.

### DIFF
--- a/README.md
+++ b/README.md
@@ -121,6 +121,17 @@ Commands:
   [h]elp                                   Show this help
 ```
 
+### screen and scrollback capture
+
+`zmx history --screen <name>` prints only the active screen, and
+`zmx history --scrollback N <name>` adds up to `N` rows of scrollback above it.
+Both work with `--vt` and `--html`. The daemon selects the rows before
+formatting them, so a tool that polls a session's current state no longer
+receives and discards the whole history. Alternate screens have no scrollback,
+so they always return just the screen. Sessions whose daemon predates this
+feature keep serving plain `history`; the scoped flags report that the daemon
+is too old.
+
 ## nested sessions
 
 Nested sessions are not supported. Inside a session `ZMX_SESSION` is set, and `attach` reads it: instead of creating another client it switches the calling terminal to the session you named.
--- a/src/completions.zig
+++ b/src/completions.zig
@@ -40,9 +40,13 @@ const bash_completions =
     \\  fi
     \\
     \\  case "$prev" in
-    \\    attach|run|send|print|write|kill|history|get|set|clear|print-env|wait|tail)
+    \\    attach|run|send|print|write|kill|get|set|clear|print-env|wait|tail)
     \\      local sessions=$(zmx list --short 2>/dev/null | tr '\n' ' ')
     \\      COMPREPLY=($(compgen -W "$sessions" -- "$cur"))
+    \\      ;;
+    \\    history)
+    \\      local sessions=$(zmx list --short 2>/dev/null | tr '\n' ' ')
+    \\      COMPREPLY=($(compgen -W "--screen --scrollback --vt --html $sessions" -- "$cur"))
     \\      ;;
     \\    completions)
     \\      COMPREPLY=($(compgen -W "bash zsh fish nu" -- "$cur"))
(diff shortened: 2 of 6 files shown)


neurosnap:
Hi thanks!

What do you think about this proposal: https://github.com/neurosnap/zmx/issues/244

This would allow agents to grab the last command ran with its output.

We would basically hook into `OSC133` to capture the prompt command and output, example:

```bash
\x1b]133;A\x07 $ cat {file} \x1b]133;B\x07 \x1b]133;C\x07
{... file output} \x1b]133;D;0\x07
```

Spec can be found here: https://gitlab.freedesktop.org/Per_Bothner/specifications/blob/master/proposals/semantic-prompts.md

maddada:
@neurosnap 
Yes, what's in that spec would be great to have.

You know much better than me but only issue I foresee with this is that a ton of the Agent CLIs are using alt now (claude, grok build, opencode, even pi has it as experimental now).

So it'll be useful for some of the tools but won't work for a bunch of them.

What I'm doing now is getting the chat transcript from the jsonl/sqlite and showing that to the user, and for whatever isn't there on desk then I'm polling the last 50 lines of scrollback (or the whole screen in the case of alt clis) while the agent is working and showing messages based on that.

Issues is there's a lot of waste due to how tail works currently hence this PR which I still would need even if that spec is implemented in the future 😅

neurosnap:
Hmm that's a really good point.  Your implementation looks pretty clean so I think we can adopt this change.  Have you been testing it against your agentic workflow?

maddada:
Yes, I've been using this a lot since the start of the day and didn't face any issues.

Here's a gist with a report from AI about the different if you care.

https://gist.github.com/maddada/b264ea4f63d1633f4641f4257e17852d

There's other optimizations later but this was the biggest one I wanted to share. Thought it might help others.

PR #257, opened 2026-09-06
Let a client claim leadership without sending input

thdxg:
## What

A new IPC `Claim` tag, and an APC sequence the client excises from stdin, letting an attached client become the session's **leader** — the one whose window size drives the pty — without sending input.

Also fixes `handleDetachAll` leaving a stale `leader_client_fd`.

## Why

zmx hands leadership to any client that sends real user input (`isUserInput`), which is the right rule for a human at a keyboard. It leaves no way for a GUI frontend showing **one session in two panes** to say "size the pty from this one" when the user clicks or focuses a pane: the only lever is typing, and the keystroke would reach their running program.

This is for Macterm's multi-window work (thdxg/macterm#345), where a mirrored pane is a second client on the same session.

## How

**APC (`ESC _ zmx;claim ESC \`)** because no keyboard produces one and terminals discard unknown APC, so a stray sequence reaching a program that predates this is inert. `ipc.Tag` is non-exhaustive by design, so an old daemon ignores the message rather than choking.

**`ClaimFilter` is deliberately not shaped like `isCtrlBackslash`.** That helper returns a bool and lets its caller drop the *entire* read — fine for a detach key, which is a keystroke and arrives alone. A claim is written programmatically and can land in the same 4096-byte read as whatever the user is typing, so an all-or-nothing match would swallow their keystrokes. The filter excises occurrences and forwards the rest.

**The carry has a floor of three bytes**, and this is the subtle part. A partial sequence can span two reads, so the filter holds a trailing prefix back — but a one- or two-byte partial is a bare `ESC` or `ESC _`, and **Escape is pressed constantly in a TUI**. Holding it until the user's next keystroke would be visible input lag. Three bytes (`ESC _ z`) is already a sequence no keyboard produces. The cost is a claim split within its first two bytes being dropped rather than rejoined — which needs the split to land exactly there, and which the next keystroke recovers from anyway.

**The `handleDetachAll` fix**: `closeClient` nils the leader per client, but that path clears the client list wholesale and left the fd behind. The next client to attach then found `leader_client_fd` non-null, so `handleInit` neither made it leader nor resized the pty — the session stayed stuck at the detached client's size until someone typed. Worse if the kernel recycled that fd number, which makes an unrelated client leader by coincidence.

## Verified

`zig build test` — 82/82 pass, including eight new `ClaimFilter` cases (excise, keystrokes preserved alongside a claim, multiple claims, split across reads, trailing-ESC not held, a held partial that turns out not to be a claim, and a bound on carry growth) and three daemon cases (claim promotes, claim by the existing leader is a no-op and sends no redundant size request, detach-all clears the leader).

Each test was checked to be non-vacuous; notably, dropping `min_carry` to 1 fails the trailing-escape test, which is what pins that design decision.

**`test/claim_leader.py`** drives two real PTYs at different sizes, proving the two properties the in-process tests cannot reach. Measured:

```
A attached at 100x30  -> shell sees 30x100
B attached at  40x12  -> shell still sees 30x100   (attach is non-disruptive)
B sends a CLAIM       -> shell sees 12x40          (no keystroke reached the program)
```

Python rather than pure bash because the test needs two PTYs at known, different window sizes, which `bats` cannot allocate. Wrapped in `test/leader.bats`, which skips when python3 is absent.

## Notes for reviewers

**Registered as a downstream patch.** Source changes on this fork survive `sync-upstream` only as a `.patch` under `.github/downstream` listed in `DOWNSTREAM_FILES`; anything else is reverted the next time main is reset to pristine upstream. Added to both the sync workflow and the drift check, and verified with `git apply --check` against the current pristine base.

**No CHANGELOG entry on purpose.** `CHANGELOG.md` is upstream's file and among the most frequently edited in any repo, so carrying it in a fork patch would make the patch a conflict magnet — and `sync-upstream` fails loudly when a patch stops applying. If this is ever proposed upstream, the changelog entry belongs in that PR.


thdxg:
Opened by mistake against upstream — this is a downstream change for the thdxg/zmx fork. Closing; apologies for the noise.

